PR #9, opened 2015-03-14
Update README.md

ocefpaf:


README.md
```diff
@@ -1,5 +1,8 @@
-# units
+# cf_units
 
-A fork of iris units module
+Units of measure.
```
rhattersley: How about: `Units of measure as required by the Climate and Forecast (CF) metadata conventions.`?

ocefpaf:
@rhattersley yep, better.

rhattersley:
At some point you might want to make your own fork of cf_units so that the branches you're submitting as PRs are your own and can't be accidentally trampled on by others.

ocefpaf:
Just did that.  The next PR should be OK.

PR #22, opened 2015-04-26
Improving python 3 compatibility

ocefpaf:
There are still two failures that needs some attention.

### DIFF
--- a/cf_units/cf_units.py
+++ b/cf_units/cf_units.py
@@ -26,7 +26,7 @@ See also: `UDUNITS-2
 """
 
 from __future__ import (absolute_import, division, print_function)
-
+from six.moves import range, zip
 import six
 
 from contextlib import contextmanager
@@ -371,7 +371,7 @@ if not _ud_system:
         if _ud_system is None:
             _alt_xml_path = os.path.join(sys.prefix, 'share',
                                          'udunits', 'udunits2.xml')
-            _ud_system = _ut_read_xml(_alt_xml_path)
+            _ud_system = _ut_read_xml(_alt_xml_path.encode())
     if not _ud_system:
         _status_msg = 'UNKNOWN'
         _error_msg = ''
@@ -1330,7 +1330,7 @@ class Unit(_OrderedHashable):
                                ctypes.sizeof(string_buffer), bitmask)
             if depth < 0:
                 self._raise_error('Failed to format %r' % self)
-        return string_buffer.value.decode()
+        return str(string_buffer.value.decode('ascii'))
 
     @property
     def name(self):
@@ -1876,8 +1876,8 @@ class Unit(_OrderedHashable):
             >>> import numpy as np
             >>> c = cf_units.Unit('deg_c')
             >>> f = cf_units.Unit('deg_f')
-            >>> print(c.convert(0, f))
-            32.0
+            >>> c.convert(0, f)
+            31.999999999999886
             >>> c.convert(0, f, cf_units.FLOAT32)
             32.0
             >>> a64 = np.arange(10, dtype=np.float64)
--- a/cf_units/cf_units.py
+++ b/cf_units/cf_units.py
@@ -27,6 +27,8 @@ See also: `UDUNITS-2
 
 from __future__ import (absolute_import, division, print_function)
 
+import six
+
 from contextlib import contextmanager
 import copy
 import ctypes
@@ -39,7 +41,7 @@ import netcdftime
 import numpy as np
 
 from . import config
-from . import util
+from .util import approx_equal
 
 
 __all__ = ['CALENDAR_STANDARD',
@@ -329,13 +331,13 @@ if _lib_ud is None:
     _ut_scale.argtypes = [ctypes.c_double, ctypes.c_void_p]
     _ut_scale.restype = ctypes.c_void_p
 
-    # convenience dictionary for the Unit convert method
+    # Convenience dictionary for the Unit convert method.
     _cv_convert_scalar = {FLOAT32: _cv_convert_float,
                           FLOAT64: _cv_convert_double}
     _cv_convert_array = {FLOAT32: _cv_convert_floats,
                          FLOAT64: _cv_convert_doubles}
     _numpy2ctypes = {np.float32: FLOAT32, np.float64: FLOAT64}
-    _ctypes2numpy = {v: k for k, v in _numpy2ctypes.iteritems()}
+    _ctypes2numpy = {v: k for k, v in _numpy2ctypes.items()}
 
 
 @contextmanager
@@ -748,7 +750,7 @@ def num2date(time_value, unit, calendar):
 ########################################################################
 
 def _Unit(category, ut_unit, calendar=None, origin=None):
-    unit = util._OrderedHashable.__new__(Unit)
+    unit = _OrderedHashable.__new__(Unit)
     unit._init(category, ut_unit, calendar, origin)
     return unit
 
@@ -769,7 +771,7 @@ def as_unit(unit):
         result = unit
     else:
         result = None
-        use_cache = isinstance(unit, basestring) or unit is None
+        use_cache = isinstance(unit, six.string_types) or unit is None
         if use_cache:
             result = _CACHE.get(unit)
         if result is None:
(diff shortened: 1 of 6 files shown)


ocefpaf:
The `TypeError` we are seeing on travis:

``` python
TypeError: unhashable type: 'Unit'
TypeError: unorderable types: Unit() < Unit()
```

makes me think that the Metaclass is not working as expected on python3,
https://github.com/SciTools/cf_units/pull/22/files#diff-2223a900628a7d260756cc9064871fa4L85
but I have no idea how to fix this :frowning:

@pelson and @rhattersley any clue about what I should do here?

rhattersley:
I'd get rid of the metaclass and move the relevant functionality down to the `Unit` class.

It still makes sense for `Unit` to be (effectively) immutable and hashable, but that doesn't need a metaclass. (NB. I know there's `collections.namedtuple()` but that would make `Unit` a subclass of `tuple` - there are times when that's annoying.)

ocefpaf:
Thanks @rhattersley!  That is something I can understand :grin: 

I will try to do that and cry for help if I find any problems.

ocefpaf:
@rhattersley do not merge this yet.  I am learning as I go here...

I migrated most of the relevant code to the `Unit` class and I think got the `immutable` and `hashable` working.

I could not get `sort` to work :sob:   `cmp` does not exist in python3 and everything I read did not work.

Also, I am unsure how to deal with the `__new__` constructor.  So I left both `_MetaOrderedHashable` and `_OrderedHashable` there temporarily until I sort this out.

cf_units/cf_units.py
```diff
@@ -830,7 +833,47 @@ class Unit(util._OrderedHashable):
...
+
+    def __ne__(self, other):
+        # Since we've defined __eq__ we should also define __ne__.
+        return not self == other
+
+    # Provide default ordering semantics
+
+    def __lt__(self, other):
+        a, b = self._identity(), other._identity()
+        return (a > b) - (a < b)
```
QuLogic: This looks like an implementation for `__cmp__`, not `__lt__`. It can return either 1, -1, or 0.

ocefpaf: @QuLogic indeed it is.  (Full disclaimer I do not know what I am doing here :stuck_out_tongue_winking_eye:)

I read a few sites mentioning that those lines are what I needed to implement `__cmp__` in python 3 while other said I needed `__lt__`.  Clearly I mixed them up and made a mess.  How you would go about implementing this?  I have no idea.

QuLogic: Well, less-than would just be `return a < b`, no?

ocefpaf: Aha, makes sense!  This is what I get for copying-and-pasting without
thinking!  Thanks!

cf_units/cf_units.py
```diff
@@ -1018,7 +1061,7 @@ def is_time(self):
         if self.is_unknown() or self.is_no_unit():
             result = False
         else:
-            day = _ut_get_unit_by_name(_ud_system, 'day')
+            day = _ut_get_unit_by_name(_ud_system, 'day'.encode())
```
QuLogic: `b'day'` and you could avoid the encoding, probably.

cf_units/cf_units.py
```diff
@@ -1376,7 +1419,7 @@ def offset_by_time(self, origin):
 
         """
 
-        if not isinstance(origin, (int, float, long)):
+        if not isinstance(origin, (int, float, six.integer_types)):
```
QuLogic: `int` is redundant here now, right?

cf_units/cf_units.py
```diff
@@ -1871,8 +1914,9 @@ def convert(self, value, other, ctype=FLOAT64):
                         if issubclass(value_copy.dtype.type, np.integer):
                             value_copy = value_copy.astype(
                                 _ctypes2numpy[ctype])
-                        # strict type check of numpy array
-                        if value_copy.dtype.type not in _numpy2ctypes.keys():
+                        # Strict type check of numpy array.
+                        keys = list(_numpy2ctypes.keys())
```
QuLogic: I don't think `list`-ification is needed for set evaluation.

jhamman: You should be able to safely do this, never creating a list of keys: 

``` python
if value_copy.dtype.type not in _numpy2ctypes:
    ....
```

same goes for line 1932.

ocefpaf:
Thanks @jhamman and @QuLogic.  This PR is not ready yet, but it is taking form!

cf_units/cf_units.py
```diff
@@ -1871,8 +1914,9 @@ def convert(self, value, other, ctype=FLOAT64):
                         if issubclass(value_copy.dtype.type, np.integer):
                             value_copy = value_copy.astype(
                                 _ctypes2numpy[ctype])
-                        # strict type check of numpy array
-                        if value_copy.dtype.type not in _numpy2ctypes.keys():
+                        # Strict type check of numpy array.
+                        keys = _numpy2ctypes.keys()
```
jhamman: This can be simplified further:

``` python

                         if value_copy.dtype.type not in _numpy2ctypes:
                             raise TypeError(
                                 "Expect a numpy array of '%s' or '%s'" %
                                 tuple(sorted(_numpy2ctypes.keys())))
```

To check if a value is a key in a dictionary, you can just do:

``` python
d = {'a': 1, 'b': 2}
'a' in d  # <-- True
# is equivalent to
'a' in d.keys()

'aa' not in d  # <-- True
# is equivalent to
'aa' not in d.keys()
```

same in line 1932.

QuLogic:
Any updates? Also, is the unit handling code in iris going to be replaced by this one? I don't want to be doing duplicate work on Python 3 stuff there.

ocefpaf:
> Any updates?

@QuLogic not really.  I am travelling now and it will take some time to get back to this.

> Also, is the unit handling code in iris going to be replaced by this one? 

I would like to, but only @rhattersley and @pelson can answer that.

>  I don't want to be doing duplicate work on Python 3 stuff there.

The work here is almost done. I am stuck to get rid of the metaclass, which should be something simple but my knowledge of metaclasses is == `None`.

If you have something that works for iris I can close this and port your work here later.

pelson:
> Also, is the unit handling code in iris going to be replaced by this one? 

That is the aim, yes.

QuLogic:
> I am stuck to get rid of the metaclass, which should be something simple but my knowledge of metaclasses is == `None`.

I don't know too much about metaclasses besides what's in the `six` documentation, but I'm pretty sure I did just about the same thing as you. I can't really say whether it's working or not as `iris` will take quite a while longer before it's ready.

ocefpaf:
> I don't know too much about metaclasses besides what's in the six documentation, but I'm pretty sure I did just about the same thing as you. I can't really say whether it's working or not as iris will take quite a while longer before it's ready.

Thanks. I will take a look at what you did once I get back in the office.

The goal here is to get rid of the metaclass by incorporating its code into
the main Unit class.

cf_units/cf_units.py
```diff
@@ -924,14 +968,14 @@ def __init__(self, unit, calendar=None):
             unit = _NO_UNIT_STRING
         else:
             category = _CATEGORY_UDUNIT
-            ut_unit = _ut_parse(_ud_system, unit, UT_ASCII)
+            ut_unit = _ut_parse(_ud_system, unit.encode(), UT_ASCII)
```
QuLogic: I think it might be safer to specify 'ascii' as the encoding here.

QuLogic:
Maybe you'll want to pull the Python 3 support directly [from iris here](https://github.com/QuLogic/iris/commits/py3k/lib/iris/unit.py) (Note: this branch is not currently the subject of a PR, and may or may not be rebased judiciously until such time that it is.)

ocefpaf:
Thanks @QuLogic!  Somehow this slipped under my radar.  I will see what I can pull from there.

ocefpaf:
@rhattersley sorry for letting this PR sit for so long.  Conference travels and day job got in the way of fun work.

> I'd get rid of the metaclass and move the relevant functionality down to the Unit class.

I cannot to do this.  It turns out my Python `metaclass` knowledge is not enough to do this correctly.

Still this PR implements Python 3.  What remains to be done is folding `_MetaOrderedHashable` and `_OrderedHashable` inthe `Unit` class and I believe we can address that in another PR.

cf_units/cf_units.py
ocefpaf: Done.  Thanks!

pelson:
:+1:

PR #37, opened 2015-11-03
New iris_pr_1813

ocefpaf:
Nothing like waiting at the airport to get some PRs out of the way...

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -57,6 +57,7 @@ __all__ = ['CALENDAR_STANDARD',
            'CALENDARS',
            'UT_NAMES',
            'UT_DEFINITION',
+           'UT_ASCII',
            'FLOAT32',
            'FLOAT64',
            'julian_day2date',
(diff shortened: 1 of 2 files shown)


cf_units/__init__.py
```diff
@@ -57,8 +57,13 @@
            'CALENDARS',
            'UT_NAMES',
            'UT_DEFINITION',
+           'UT_ASCII',
            'FLOAT32',
            'FLOAT64',
+           '_NO_UNIT',
+           '_NO_UNIT_STRING',
+           '_NO_UNIT_SYMBOL',
+           '_UNKNOWN_UNIT_SYMBOL',
```
DPeterK: What is the benefit of adding these to the `__all__`? It's a bit unusual to add "private" objects to the `__all__` like this...

ocefpaf: No benefit and no reason.  Don't know why copied those there in the first place.

I removed those and rebased.

PR #39, opened 2015-11-04
Python 3.5 testing

ocefpaf:
We need to `pip install pep8==1.5.7` because that version does not exist on conda+python35.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 python:
   - 2.7
   - 3.4
+  - 3.5
 
 install:
   # Install miniconda
@@ -25,12 +26,12 @@ install:
   - conda create --quiet -n $ENV_NAME python=$TRAVIS_PYTHON_VERSION
   - source activate $ENV_NAME
 
-  # Customise the testing environment
-  # ---------------------------------
+  # Customise the testing environment.
+  # ----------------------------------
   - conda config --add channels scitools
   - conda install --quiet --file conda-requirements.txt
 
-  # Output debug info
+  # Output debug info.
   - conda list
   - conda info -a
 
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -42,7 +42,7 @@ import netcdftime
 import numpy as np
 
 from . import config
-from .util import approx_equal
+from .util import _OrderedHashable, approx_equal
 
 __version__ = '0.1.1'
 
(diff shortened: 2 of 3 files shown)


ajdawson:
What problems do later PEP8 versions bring, I forgot? I was just wondering if it is feasible to actually use a newer version instead of continue to pin?

DPeterK:
@ajdawson the changelog can be found partway down the [pypi page for pep8](https://pypi.python.org/pypi/pep8). Quite a few things have changed (too many to summarise here) and there is no real theme to the changes either.

ajdawson:
I know that, I just wanted a recap of what _specific_ difficulties it brings for _this_ project. Are we just expecting to use pep8 v1.5.7 forever? Sounds like a potentially bad plan, and now might be a good time to think about how to deal with it.

DPeterK:
> specific difficulties it brings for this project

Sorry, my bad.

Clearly staying with pep8 v1.5.7 until the end of time is not a great idea, but if we do change versions I'd say we should at least consider changing all major SciTools repos at the same time so that we don't effectively have different coding standards expectations across the repos.

Of course, we could leave pep8 in favour of a different tool - [flake8](https://pypi.python.org/pypi/flake8) perhaps. This is already making an implicit appearance in our code with the `# noqa` comments that accompany `six` imports.

ocefpaf:
@ajdawson and @dkillick it turns out that this project did have a pep8 failure :flushed: 

Maybe iris needs a more careful consideration to move away from `pep8 v1.5.7`, but `cf_units` is OK now.

PS: The pep8 failure was due to a remnant of my failed attempts to address this [comment](https://github.com/SciTools/cf_units/pull/22#issuecomment-96552912).  That needs the attention of a real developer!  Metaclasses are not my forte.

ajdawson:
OK, so it is hard to ditch 1.5.7 right now, no problem. Instead of a requirements file could be use an environment.yaml file (see http://conda.pydata.org/docs/using/envs.html#share-an-environment)? This can encode both conda package and pip dependencies all in one place which is nice. I guess the downside is it will probably encode all versions of all packages which may not be OK. May be worth a little moire investigation anyway.

ocefpaf:
> OK, so it is hard to ditch 1.5.7 right now.

Just to be clear. `cf_units` **is** OK with latest pep8.

I fixed the remaining pep8 infringement (that I introduced some time ago), removed the pep8 version requirement and rebased.

(It might be harder to ditch it for the rest of iris code though.)

rhattersley:
:+1: for embracing the future :rocket: :grin:

PR #42, opened 2015-11-19
Bake udunits version into package name

ocefpaf:
I used the same branch name and title to make @rhattersley happy :smile_cat:


rhattersley:
> I used the same branch name and title to make @rhattersley happy :smile_cat:

:stuck_out_tongue:

PR #55, opened 2016-05-04
Split requirements and added to setup.py

### LINKED ISSUE #54: Package dependencies
I've been troubleshooting a bunch of tox issues for a project that uses cf_units and I found where the source of the trouble is. cf_units defines the project dependencies in a conda requirements file but not in the setup.py file

Specifically the dependency that's giving me grief is six.

Could you also include the project dependencies in the setup.py file? It might be as simple as

``` python
install_requires = [line.strip() for line in open('conda_requirements.txt')]
```

ocefpaf:
This change makes `cf_units` more `pip` friendly.

Closes #54

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,9 @@ long_description = '{}'.format(read('README.rst'))
 cmdclass = {'test': PyTest}
 cmdclass.update(versioneer.get_cmdclass())
 
+with open('requirements.txt') as f:
+    require = f.readlines()
+install_requires = [r.strip() for r in require]
 
 setup(
     name='cf_units',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -41,6 +41,9 @@ def read(*parts):
 cmdclass = {'test': PyTest}
 cmdclass.update(versioneer.get_cmdclass())
 
+with open('requirements.txt') as f:
+    require = f.readlines()
+install_requires = [r.strip() for r in require]
```
rhattersley, line 46: Happily, setuptools already ignores entries with a leading `#` so there's no need to filter them out here.

ocefpaf:
Thanks @rhattersley!

PR #56, opened 2016-05-05
Warn if udunits2 fails

ocefpaf:
Solves https://github.com/ioos/compliance-checker/issues/265#issuecomment-215791057

Ping @lukecampbell

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -220,10 +220,15 @@ if _lib_c is None:
 # load the libudunits2 shared library
 #
 if _lib_ud is None:
-    _lib_ud = config.get_option(
+    _udunits2_c = config.get_option(
         'System', 'udunits2_path',
         default=ctypes.util.find_library('udunits2'))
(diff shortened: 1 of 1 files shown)


cf_units/__init__.py
```diff
@@ -220,10 +220,15 @@
 # load the libudunits2 shared library
 #
 if _lib_ud is None:
-    _lib_ud = config.get_option(
+    udunits2_c = config.get_option(
```
rhattersley: The convention so far has been for these kinds of helper variables to have an underscore prefix to mark them as private and stop them from turning up in tab-completion.

ocefpaf: Thanks @rhattersley. I just fixed that.

rhattersley:
Thanks @ajdawson :+1:

ocefpaf:
Thanks for merging @rhattersley. Can you take a look at #54 too. It is also minor changes.

PR #60, opened 2016-06-23
Converted README to md

DPeterK:
Converted the README file to an md file so that the badges work nicely. I've also taken then opportunity to polish and improve the contents of the README a little, and corrected a link that wouldn't work in an md file.

(@bjlittle sorry! I got there first :stuck_out_tongue:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ cf_units is licenced under the GNU Lesser General Public License (LGPLv3).
 The full text of the licence can be found in the COPYING and COPYING.LESSER
 files.
 
-Development occurs at https://github.com/SciTools/cf_units.
+Development occurs at https://github.com/SciTools/cf_units. Documentation for
+cf_units can be found at http://scitools.org.uk/cf_units/index.html.
 
 
 See also:
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ cf\_units
 
 \(C) British Crown Copyright 2015 - 2016, Met Office
 
+Introduction
+------------
 Units of measure as required by the Climate and Forecast (CF) metadata
 conventions.
 
@@ -18,9 +20,9 @@ cf_units is licenced under the GNU Lesser General Public License (LGPLv3).
 The full text of the licence can be found in the COPYING and COPYING.LESSER
 files.
 
-Development occurs at https://github.com/SciTools/cf_units. Documentation for
-cf_units can be found at http://scitools.org.uk/cf_units/index.html.
+See also
+--------
 
-
-See also:
-`UDUNITS-2 <http://www.unidata.ucar.edu/software/udunits/udunits-2/udunits2.html>`__
+* Development occurs at https://github.com/SciTools/cf_units.
+* Documentation for cf_units can be found at http://scitools.org.uk/cf_units/index.html.
+* The library wrapped by cf_units: [UDUNITS-2](http://www.unidata.ucar.edu/software/udunits/udunits-2/udunits2.html).


ajdawson:
Why wouldn't you just change the link syntax to use rst style? The headings
you added are rst style anyway...

DPeterK:
@ajdawson the simpler change was just to convert to markdown...

ajdawson:
I fail to see how it is simpler to change the format of the file than to just use the correct syntax. Markdown isn't one of the supported formats for setuptools, whereas rst is, so now we'll get a warning when when building a source distribution with setuptools:

```
warning: cmd_sdist: standard file not found: should have one of README, README.rst, README.txt
```

The link syntax would simply have been:

```
.. image:: https://travis-ci.org/scitools/cf_units.svg?branch=master
   :target: https://travis-ci.org/scitools/cf_units
```

README.md
```diff
@@ -5,6 +5,8 @@ cf\_units
 
 \(C) British Crown Copyright 2015 - 2016, Met Office
 
+Introduction
+------------
```
ajdawson, line 9: This is an RST style heading, markdown headings use the `#`, see https://help.github.com/articles/basic-writing-and-formatting-syntax/#headings

README.md
```diff
@@ -18,8 +20,9 @@ cf_units is licenced under the GNU Lesser General Public License (LGPLv3).
 The full text of the licence can be found in the COPYING and COPYING.LESSER
 files.
 
-Development occurs at https://github.com/SciTools/cf_units.
+See also
+--------
```
ajdawson, line 24: Same here.

DPeterK:
And yet, looking around SciTools shows an over 90% lean toward using markdown and _not_ rst for the library's readme. If this warning wasn't a problem before with any of those repos, why make it a problem now with this repo?

README.md
DPeterK, line 9: And yet the headings are all rendered correctly as top level or first sub-level headings...

Also, I quite like the look of the horizontal rules under each sub-heading.

bjlittle, line 9: @ajdawson What about this, see https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#headers, and https://daringfireball.net/projects/markdown/basics

ajdawson:
My point is that this seems like an unnecessary backward step. Sure, things render OK with those heading styles, and sure we use markdown in other places, but we already had an rst README, and all it needed was a bit of care to use the correct syntax. In the end it doesn't really matter too much, but it just seems mad to make this change when the alternative (using the correct syntax) was so simple and within reach.

ajdawson:
> If this warning wasn't a problem before with any of those repos, why make it a problem now with this repo?

This repo is not the others, it is its own thing, and it didn't previously generate that warning on build and now it does. We should be aiming to look forward, not comforting ourselves that everything is fine because we've done it like this before.

DPeterK:
@ajdawson perhaps we haven't been as clear as we could have on one of our key drivers for converting to MD: you can easily line up the badges on a single line. This is apparently (quote @bjlittle) almost impossible to achieve using RST. (That said, if you can convince RST to play and look nice then we will happily choose that change over this one...)

In light of this it seemed a no-brainer to just convert to MD - so far as we could see RST and MD were pretty much interchangeable, and MD is evidently github's preferred markdown language. But as you have pointed out that setuptools doesn't support MD that needs to be considered, and honestly our consideration is that it isn't worth fussing over a finicky warning in setuptools, dare I say it especially as we are using setuptools less and less...

ajdawson:
> one of our key drivers for converting to MD: you can easily line up the badges on a single line. This is apparently (quote @bjlittle) almost impossible to achieve using RST. (That said, if you can convince RST to play and look nice then we will happily choose that change over this one...)

See one (or more) of the following:

https://github.com/pydata/xarray/blob/master/README.rst
https://github.com/conda/conda/blob/master/README.rst
https://github.com/dask/dask/blob/master/README.rst

Seems like it should be do-able.

bjlittle:
@ajdawson I'm going to merge this, then convert back to .rst (just to capture @dkillick changes)

PR #64, opened 2016-07-02
Setup tweaks

QuLogic:
Two small things: use context managers for I/O + a fixed encoding. Secondly, change install path for license. That path is relative to `sys.prefix`, meaning it goes in `/usr/cf_units` or `~/.local/cf_units`, which is a rather odd place. It now goes in `/usr/share/doc/cf_units` or `~/.local/share/doc/cf_units`, which is a bit more reasonable.


bjlittle:
@QuLogic Great, thanks! :+1:

PR #65, opened 2016-07-28
On Windows, Temporarily change current working directory

ghost:
to the system drive before loading udunits2.xml

Fixes: https://github.com/SciTools/conda-recipes-scitools/issues/115#issuecomment-146829335

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -385,13 +385,15 @@ if not _ud_system:
                 default=os.path.join(sys.prefix, 'share', 'udunits',
                                      'udunits2.xml'))
             if os.name == 'nt':
-                # On Windows systems we need to read the xml while on the system drive
-                # but we don't want to change the working directory every time someone loads this package
+                # On Windows systems we need to read the xml while on the
+                # system drive but we don't want to change the
+                # working directory every time someone loads this package,
                 # so we will restore the original working directory
                 cur_wd = os.getcwd()
                 system_drive = os.getenv("SystemDrive")
                 if not system_drive:
-                    # if for some reason this system doesn't have a SystemDrive env variable C: is a safe assumption
+                    # if for some reason this system doesn't have a
+                    # SystemDrive env variable C: is a safe assumption
                     system_drive = "C:"
                 os.chdir(system_drive)
                 _ud_system = _ut_read_xml(_alt_xml_path.encode())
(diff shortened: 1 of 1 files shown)


cf_units/__init__.py
```diff
@@ -384,7 +384,22 @@ def suppress_errors():
...
+            if os.name == 'nt':
+                # On Windows systems we need to read the xml while on the
+                # system drive but we don't want to change the
+                # working directory every time someone loads this package,
+                # so we will restore the original working directory
+                cur_wd = os.getcwd()
+                system_drive = os.getenv("SystemDrive")
+                if not system_drive:
+                    # if for some reason this system doesn't have a
+                    # SystemDrive env variable C: is a safe assumption
```
DPeterK, line 396: I don't believe this is a safe assumption to make...

cf_units/__init__.py
```diff
@@ -384,7 +384,22 @@ def suppress_errors():
                 'System', 'udunits2_xml_path',
                 default=os.path.join(sys.prefix, 'share', 'udunits',
                                      'udunits2.xml'))
-            _ud_system = _ut_read_xml(_alt_xml_path.encode())
+            if os.name == 'nt':
+                # On Windows systems we need to read the xml while on the
+                # system drive but we don't want to change the
```
DPeterK, line 389: This assertion is actually incorrect: you need to read the XML while your CWD is **on the drive that the XML is located upon**.

DPeterK:
Much as it is good to see some progress with trying to fix this annoying problem, I disagree with the approach suggested. The approach really is a hack, and in fact any fix for this within cf_units will be a hack, because the problem lies in udunits, not cf_units.

For reference, we have an open issue with UDUNITS-2 about this: https://github.com/Unidata/UDUNITS-2/issues/43.

cf_units/__init__.py
ghost, line 389: Well thanks for considering anyway.  I agree this solution is a bit hacky,
and will look forward to it getting fixed in udunits someday.

Colin Talbert
GIS Analyst and Developer
US Geological Survey

talbertc@usgs.gov

USGS Fort Collins Science Center
2150 Centre Ave. Bldg. C
Fort Collins, CO 80526
(970) 226-9425

USGS North Central Climate Science Center
(970) 492-4283

Work schedule:
Monday       - 7:00 - 3:00  (FORT)
Tuesday       -  7:00 - 3:00  (NC CSC)
Wednesday  -  7:00 - 3:00 (FORT)
Thursday      -  7:00 - 5:00 (FORT)
Friday          -  7:00 - 5:00 (FORT)

pelson:
Looks like this may have been solved in the upstream issue. @dkillick / @ColinTalbert, are you able to confirm with the latest udunits2 and cf_units?

PR #66, opened 2016-09-12
Ensure dates are rounded to nearest second

djkirkham:
Ensures `num2date` returns datetimes rounded to one second. Allowing microsecond precision introduces rounding errors.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -694,16 +694,12 @@ def _discard_microsecond(date):
     # Return a date with the microsecond componenet discarded. `date` should be
     # a datetime.datetime or netcdftime.datetime object, or a list of such
     # objects
-    try:
-        l = list(date)
-    except TypeError:
-        dates = np.array([date])
+    is_scalar = False
+    if not hasattr(date, '__iter__'):
+        date = [date]
         is_scalar = True
-    else:
-        dates = np.array(l)
-        is_scalar = False
     dates = [d.__class__(d.year, d.month, d.day, d.hour, d.minute, d.second)
-             for d in dates]
+             for d in date]
     return dates[0] if is_scalar else dates
 
 
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -753,7 +753,41 @@ def num2date(time_value, unit, calendar):
     if unit_string.endswith(" since epoch"):
         unit_string = unit_string.replace("epoch", EPOCH)
     cdftime = netcdftime.utime(unit_string, calendar=calendar)
-    return cdftime.num2date(time_value)
+    return _num2date_to_nearest_second(time_value, cdftime)
+
+
+def _num2date_to_nearest_second(time_value, utime):
+    # Return datetime encoding of numeric time value with respect to the given
+    # time reference units, with a resolution of 1 second.
+
+    # We account for the edge case where the time is in seconds and has a
+    # half second: utime.num2date() may produce a date that would round
+    # down.
+    #
+    # Note that this behaviour is different to the num2date function in older
+    # versions of netcdftime that didn't have microsecond precision. In those
+    # versions, a half-second value would be rounded up or down arbitrarily. It
+    # is probably not possible to replicate that behaviour with the current
+    # version (1.4.1), if one wished to do so for the sake of consistency.
+    has_half_second = utime.units == 'seconds' and \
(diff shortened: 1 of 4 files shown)


cf_units/__init__.py
```diff
@@ -753,7 +753,41 @@ def num2date(time_value, unit, calendar):
...
+
+    # We account for the edge case where the time is in seconds and has a
+    # half second: utime.num2date() may produce a date that would round
+    # down.
+    #
+    # Note that this behaviour is different to the num2date function in older
+    # versions of netcdftime that didn't have microsecond precision. In those
+    # versions, a half-second value would be rounded up or down arbitrarily. It
+    # is probably not possible to replicate that behaviour with the current
+    # version (1.4.1), if one wished to do so for the sake of consistency.
```
DPeterK, line 815: Out of interest, why has this been provided as a multi-line comment rather than a docstring?

djkirkham: This is just a comment for developers, so I don't think it should go in a docstring. I can provide a docstring based on L760-761 if you like, although this is a private function.

djkirkham:
Hmm... previously the `num2date` functions would have worked when passing a list of values, but after my change they no longer do. It's not in the docstring, but I think I should make it work again. Should I change the doctring while I'm at it?

DPeterK:
> Should I change the doctring

Explicit is good :+1:

DPeterK:
Have you investigated whether you need to do the same for `date2num`? That is, has the accuracy of that function also been affected by the microseconds change?

marqh:
I support this in principal.

I wonder whether we want to publicise this functionality and whether we want to enable access to microseconds

@djkirkham do you feel that implementing such changes would be simply additive, from the changes posted here?

djkirkham:
@dkillick Good point. I hadn't looked into it, and it turns out that behaviour has changed. In the older version of netcdftime it seems `date2num` discards the microsecond component, but now it doesn't. I suppose we should make changes to keep this behaviour too.

djkirkham:
@marqh 

> I wonder whether we want to publicise this functionality

Do you mean the functionality of being able to pass a list as argument?

> do you feel that implementing such changes would be simply additive, from the changes posted here?

It should be simple to add an optional argument to keep the microseconds

DPeterK:
> I suppose we should make changes to keep this behaviour too.

We should indeed! And we should add a test that proves that the behaviour is correct, given that it's slipped through up to now.

djkirkham:
Actually I'm not sure about including microseconds, since they're not available in the older version of netcdftime. Should this be included as a note in the docstring? What should happen if the user requests microseconds when using the older version? Raise an exception? Ignore it and return seconds resolution?

marqh:
> Actually I'm not sure about including microseconds, since they're not available in the older version of netcdftime. Should this be included as a note in the docstring? What should happen if the user requests microseconds when using the older version? Raise an exception? Ignore it and return seconds resolution?

This is a good point.

for now, I suggest we maintain the current behaviour, which explicitly ignores microseconds for all versions on netcdf4-python (this PR, as I understand it)

Changes to include these, to provide options to functions and to provide public access to private functions are easy to add at a later date

DPeterK:
> for now, I suggest we maintain the current behaviour

I agree. Banking this will be an excellent step forward and then we can take the time to think about the more complicated questions that @djkirkham raised.

cf_units/__init__.py
```diff
@@ -753,7 +770,50 @@ def num2date(time_value, unit, calendar):
     if unit_string.endswith(" since epoch"):
         unit_string = unit_string.replace("epoch", EPOCH)
     cdftime = netcdftime.utime(unit_string, calendar=calendar)
-    return cdftime.num2date(time_value)
+    return _num2date_to_nearest_second(time_value, cdftime)
+
+
+def _num2date_to_nearest_second(time_value, utime):
+    # Return datetime encoding of numeric time value with respect to the given
```
marqh: please make this a docstring

cf_units/__init__.py
```diff
@@ -753,7 +770,50 @@ def num2date(time_value, unit, calendar):
...
+    # time reference units, with a resolution of 1 second.
+    try:
```
marqh: I prefer the pattern used above: testing for `__iter__` to this try except pattern

PR #68, opened 2016-09-15
Fix num2date / date2num multidimensional arrays. Remove iterator support. Fix masked arrays

djkirkham:
Fixes a bug where these functions wouldn't work with multidimensional arrays following a recent change.

I've also removed support for passing iterators to these functions to keep the code clean. Iterators weren't supported before the previous change.


djkirkham:
Previously the `num2date` functions would return `None` in an maked array where the input was masked. The `date2num` function would ignore the mask. I have maintained this functionality

PR #79, opened 2017-04-11
Use self.symbol as consistent string representation

tv3141:
Currently, the string representation uses `self.origin` - the string used to create the instance. After an arithmetic operation, the returned instance does not have this attribute any more and uses `self.name`, which is most likely different from `self.origin`. `self.name` also spells out the base units which is very verbose.

I suggest using `self.symbol` as string representation:

```python
import cf_units

watt = cf_units.Unit('kg meter2 second-3')
print(watt)             # kg meter*meter/(second*second*second)
print(watt.symbol)      # W

another_watt = watt * 1
print(another_watt)            # watt -> changed
print(another_watt.symbol)     # W

watt_meter = watt * cf_units.Unit('meter')
print(watt_meter)            # meter^3-kilogram-second^-3  -> verbose
print(watt_meter.symbol)     # m3.kg.s-3
```

**Before merging: Further changes in doc strings might be necessary.**

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1670,12 +1670,12 @@ class Unit(_OrderedHashable):
         For example:
 
             >>> import cf_units
-            >>> u = cf_units.Unit('meters')
+            >>> u = cf_units.Unit('kg meter2 second-3')
             >>> str(u)
-            'meters'
+            'W'
 
         """
-        return self.origin or self.name
+        return self.symbol
 
     def __repr__(self):
         """
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1670,12 +1670,12 @@ class Unit(_OrderedHashable):
         For example:
 
             >>> import cf_units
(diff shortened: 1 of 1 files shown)


djkirkham:
Hi @tv3141, thanks for submitting this PR.

The origin string can give information about the units that is lost when using `self.symbol`. For example, the units for humidity are 'kg/kg' or 'g/kg', but using `self.symbol` would just give '1' or '0.001 1'.

Also, the origin string can be more understandable for non-SI units. When using `self.symbol`, 'miles hour-1' becomes '0.44704 m.s-1'.

I agree that `self.name` is verbose. Perhaps we could use `self.origin or self.symbol` instead.

tv3141:
I see the value of `self.origin`in special cases, but thought consistent general behavior, i.e. the unit string does not change, unless the unit changes, would outweigh this.

I think using `self.origin or self.symbol` would be an improvement. The string representation of a unit can then always be used for human consumption, such as in a plot.

PR #91, opened 2018-05-02
Fix travis ci

bjlittle:
This PR fixes the currently broken `travis-ci` tests.

I decided to flip back to `unittest` from `pytest`, simply because I couldn't get it to work using `pytest` (as it stood)

I also stopped `travis-ci` from pulling `udunits2` from `scitools` :scream: , in preference from `defaults`. Moving to a more recent version of `udunits2` i.e. `2.2.20` from `scitools` to `2.2.25` from `anaconda` highlighted that we had broken tests.

The tests were attempting to perform cross calendar `datetime` comparisions, which you can't do now (and I didn't know that you could ever do that)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,16 @@ language: python
 
 python:
   - 2.7
-  - 3.4
-  - 3.5
+  - 3.6
 
 install:
   # Install miniconda
   # -----------------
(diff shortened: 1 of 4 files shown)


bjlittle:
Ping @pelson ... fancy reviewing this?

The `travis-ci` for `cf-units` needed some love... :heart:

PR #116, opened 2018-08-05
Test on python 3.7...

ocefpaf:
... and fix the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
 
-python:
-  - 2.7
-  - 3.6
-  - 3.7
+matrix:
+  include:
+  - env: PY=2.7
+  - env: PY=3.6
+  - env: PY=3.7
 
 install:
   # Install miniconda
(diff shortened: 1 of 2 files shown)


ocefpaf:
@bjlittle or @pelson can you take a look at this PR? I kind of have use for this in CIs downstream...

lbdreyer:
@ocefpaf thanks for this! I was just about to look into adding python 3.7 but found you had already done it!

ocefpaf:
If you agree I can put together a new release with the Python 3.7 support.

lbdreyer:
@ocefpaf I think that's  a good idea :+1: 

Then it will make it easier to update all the packages that depend on cf-units to include Python 3.7 testing

ocefpaf:
@lbdreyer actually I was mistaken and only a test needed fixing. (For some reason I thought that I also modified the core code base.)

So no new releases is actually needed to run this on Python 3.7. We can wait for more functional changes to do that.

PR #120, opened 2018-09-26
Add coverage of cython

lbdreyer:
Extends the code coverage testing to include cython files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,11 @@ else:
     extra_extension_args = dict(
         runtime_library_dirs=library_dirs)
 
+# Enable cython coverage
+cython_coverage_enabled = os.environ.get('CYTHON_COVERAGE', None)
+if cython_coverage_enabled:
+    extra_extension_args.update({'define_macros': [('CYTHON_TRACE_NOGIL', '1')]})
+
 ext = 'pyx' if cythonize else 'c'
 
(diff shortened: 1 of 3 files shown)


lbdreyer:
Not sure why this is failing :confused: it may be due to the coverage going down??

lbdreyer:
This https://github.com/lemurheavy/coveralls-public/issues/565#issuecomment-120924694 suggests that we can set the tolerance to be something to allow for occasional decreases in code coverage. Unfortunately, I am not an owner so I cannot see the settings to change it.

@pelson @bjlittle I suspect one of you may an owner of this repo. Could you change the coverage tolerance setting https://coveralls.io/github/SciTools/cf-units?

setup.py
```diff
@@ -55,7 +60,7 @@ def read(*parts):
                         **extra_extension_args)
 
 if cythonize:
-    [udunits_ext] = cythonize(udunits_ext)
+    [udunits_ext] = cythonize(udunits_ext, compiler_directives={'linetrace': True})
```
lbdreyer: We could also add the `binding=True` directive?

lbdreyer, line 63: We could also add the directive: `binding=True`. That can sometime help with cython code coverage.

bjlittle:
@lbdreyer Closed by #126, thanks

PR #122, opened 2018-11-09
update INSTALL

bjlittle:
Refresh the `INSTALL` instructions.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -9,7 +9,7 @@ cf-units is available using conda for the following platforms:
  * Windows 32-bit and 64-bit.
 
 To install cf-units using conda, you must first download and install conda,
-for example from http://conda.pydata.org/miniconda.html.
+for example from https://conda.io/miniconda.html.
 
 Once conda is installed, you can install cf-units using conda on any platform with
 the following command::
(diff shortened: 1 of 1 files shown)


bjlittle:
Closes https://github.com/SciTools/cf-units/issues/117

bjlittle:
Happy to rebase once https://github.com/SciTools/cf-units/pull/123 is merged...

INSTALL
```diff
@@ -9,27 +9,27 @@ cf-units is available using conda for the following platforms:
...
+for example from https://conda.io/miniconda.html.
 
 Once conda is installed, you can install cf-units using conda on any platform with
 the following command::
 
   conda install -c conda-forge cf-units
 
 Further documentation on using conda and the features it provides can be found
-at http://conda.pydata.org/docs/intro.html.
+at http://conda.io/docs/.
```
DPeterK: `http` --> `https`

INSTALL
```diff
@@ -55,34 +55,40 @@ If you are installing dependencies with a package manager on Linux,
 you may need to install the development packages (look for a "-dev"
 postfix) in addition to the core packages.
 
-python 2.7 or later (http://www.python.org/)
-    cf-units requires Python 2.7 or later, but is not currently compatible with
-    Python 3.
+python 2.7 or later (https://www.python.org/)
+    cf-units requires Python 2.7 or later.
```
DPeterK: This line feels tautological now – it's not saying anything that hasn't already been said on the line above. If we wanted to be explicit and say something like `cf-units is compatible with Python 2 and Python 3`, but, really, that's also a given...

DPeterK (commented):
@bjlittle I think there are a couple of bits that could do with a little tidy. Otherwise it's a nice update 👍

pelson (approved):
@dkillick has this one under control 👍

bjlittle:
@dkillick Rebased and addressed your review actions. As agreed, running with this as is, thanks!

PR #127, opened 2018-11-24
flake8 + stickler

bjlittle:
This PR flake8's the code and turns on `stickler-ci`.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -719,9 +719,6 @@ class Unit(_OrderedHashable):
         for name, value in zip(self._names, values):
             object.__setattr__(self, name, value)
 
-    def _as_tuple(self):
-        return tuple(getattr(self, name) for name in self._names)
-
     # Provide hash semantics
 
     def _identity(self):
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -34,7 +34,6 @@ from contextlib import contextmanager
 import copy
 import os.path
 import sys
-import warnings
 
 import cftime
 import numpy as np
@@ -726,7 +725,7 @@ class Unit(_OrderedHashable):
     # Provide hash semantics
 
     def _identity(self):
-        return self._as_tuple()
+        return (self.name, self.calendar)
 
     def __hash__(self):
         return hash(self._identity())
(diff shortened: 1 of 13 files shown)


bjlittle:
@dkillick Got time for a review? 😸

cf_units/__init__.py
```diff
@@ -726,7 +725,7 @@ def _as_tuple(self):
     # Provide hash semantics
 
     def _identity(self):
-        return self._as_tuple()
```
DPeterK, line 729: Now that you're not calling this private method it's not used anywhere within this class (or indeed the codebase). Given that it's private, could we consider removing it as well?

It looks like this identity business is a bit of a mess! I guess that the last defined method of identical name on a class is always the one that will be run when called? If it is then all's well as we haven't changed the behaviour here. Given, however, that there don't look to be any tests that check this behaviour, could you add a test for this and confirm that the behaviour hasn't changed with this? Be good to confirm there isn't a behaviour change here.

DPeterK (commented):
@bjlittle thanks for getting the duster out! There's one thing I'd like some extra certainty on to make sure there isn't a behaviour change introduced here, but otherwise I think we're all good 👍

cf_units/__init__.py
bjlittle: `flake8` uncovered this... that's why we should use it more!

It seems that the last duplicate method was the one that was being used. This was proved by removing the duplicate and calling `self._as_tuple` version instead, which broke tests.

So I'm happy that there is test coverage that captures a change in behaviour. Happy to add an explicit test for `_identify` if you want?

The only thing that I'm not sure about is whether the `_as_tuple` method is used as part of some sort of "private" protocol, and used by an external third party package. I guess we could just remove it and if issues ensue, reinstate it with a better understanding of why it actually exists.

DPeterK: > flake8 uncovered this... that's why we should use it more!

👍 👏 

> calling `self._as_tuple` instead, which broke tests

Well, that's anecdotally evidence enough that there is sufficient, behaviour-appropriate testing in place then!

Thinking about `_as_tuple`, I think I'm inclined to say remove it. It's no longer used by the class that it's a private method of, so nothing else should be using it. Given also that it's private we aren't breaking API by removing it without warning. Think this will be good to go once that method is gone!

bjlittle:
@dkillick I've remove the unused `_as_tuple` method... looks good to me :smile:

DPeterK:
🎉 Thanks @bjlittle!

bjlittle:
Hoo-rah! Awesome, thanks @dkillick

PR #141, opened 2019-01-29
Add units to latex functionality

bjlittle:
Adds ability to render latex from the udunits2 grammar.

### DIFF
--- a/cf_units/_udunits2_parser/graph.py
+++ b/cf_units/_udunits2_parser/graph.py
@@ -109,3 +109,35 @@ class Timestamp(Terminal):
     # This is likely to change in the future, but there are some
     # gnarly test cases, and should not be undertaken lightly.
     pass
+
+
+class Visitor:
+    """
+    This class may be used to help traversing an expression graph.
+
+    It follows the same pattern as the Python ``ast.NodeVisitor``.
+    Users should typically not need to override either ``visit`` or
+    ``generic_visit``, and should instead implement ``visit_<ClassName>``.
+
+    Remember that each method MUST call ``generic_visit(node)`` if a node's
+    children should be processed.
+
+    This class is used in cf_units.latex to generate a latex representation of
+    an expression graph.
+
+    """
+    def visit(self, node):
+        """Visit a node."""
+        method = 'visit_' + node.__class__.__name__
+        visitor = getattr(self, method, self.generic_visit)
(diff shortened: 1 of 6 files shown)


cf_units/_udunits2_parser/graph.py
```diff
@@ -109,3 +109,35 @@ class Timestamp(Terminal):
     # This is likely to change in the future, but there are some
     # gnarly test cases, and should not be undertaken lightly.
     pass
+
+
+class Visitor:
```
pelson, line 114: Nice pattern 😜

cf_units/_udunits2_parser/graph.py
```diff
@@ -109,3 +109,35 @@ class Timestamp(Terminal):
...
+    """
+    This class may be used to help traversing an expression graph.
+
+    It follows the same pattern as the Python ``ast.NodeVisitor``.
+    Users should typically not need to override either ``visit`` or
+    ``generic_visit``, and should instead implement ``visit_<ClassName>``.
+
+    Remember that each method MUST call ``generic_visit(node)`` if a node's
+    children should be processed.
```
pelson: Perhaps add "alternatively, each of the children can be visited individually using the visit method (e.g. ``visit(node.lhs)``)"?

cf_units/latex.py
```diff
@@ -0,0 +1,56 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with cf-units.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import cf_units._udunits2_parser.graph as graph
+    from cf_units._udunits2_parser import parse as _parse
+except SyntaxError:
```
pelson: I'm tempted to suggest an if statement with ``six.PY2`` in it. That is going to be the surest way of us finding this when we make the code py3 only.

pelson (approved):
LGTM 👍

pelson:
Amazeballs! ⚽️

Super excited about this. It closes the loop on some *very* long standing requests to turn units into a LaTeX representation, which was most recently stoked by @pdearnshaw in https://github.com/SciTools/cf-units/pull/121#issuecomment-458604654.

tex

Issue #149, opened 2019-03-29
Comparison of values with Unit should not throw ValueError.


mrshannon:
The `__eq__` method of `Unit` attempts to convert any value it is compared with to a `Unit`.  The problem is when this fails it should return `NotImplemented` (or raise `NotImplementedError`) instead of raising `ValueError`.  Because of this, comparisons such as:

```python
Unit('meter') == {}
```
raises
```python
ValueError: [UT_UNKNOWN] Failed to parse unit "{}": b'Resource temporarily unavailable'
```
when it should return `False`.  By returning `NotImplemented` you can punt the decision to the other object or fall back on object identity (if neither object can handle the comparison).

bjlittle:
Closed by https://github.com/SciTools/cf-units/pull/150

PR #150, opened 2019-03-29
Deligate equality to other if conversion to Unit fails.

mrshannon:
This fixes issue #149 by returning `NotImplemented` if `other` is not or cannot be converted to a `Unit`.  By doing this `Unit` allows `other` to make the decision or in the case that neither can determine equality Python will fallback on identity.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1731,7 +1731,10 @@ class Unit(_OrderedHashable):
             True
 
         """
-        other = as_unit(other)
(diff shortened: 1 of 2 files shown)


bjlittle:
@mrshannon Sorry for the very tardy response... I clearly missed this, but let's bank it now (finally) and include it in a patch release

bjlittle:
@mrshannon LGTM :+1:

Issue #152, opened 2019-07-22
numpy 1.17.0 has dropped support for Python 2, causing installation to fail due to unpinned numpy requirement


ghost:
Not sure where you're at in terms of deciding to finally end support for Python 2, but I have been getting this issue since numpy released some prerelease versions of 1.17.0.

The root cause is really the unpinned dependency on numpy (which means it will just break for external reasons), but also the fact that it is installing RC versions of numpy in the first place. The only thing I can guess at this point is that the versions installed as a "setup requirement" are not correctly excluding RC versions and installing the latest production release.

Note that the 1.17.0 are flagged as prerelease on https://pypi.org/project/numpy/#history, so *shouldn't* be installed as a dependency of a production package, but for some reason they are.

It should be easy to replicate by simply trying to install cf-units into an empty Python 2 virtualenv, e.g.

```bash
$ virtualenv venv && source venv/bin/activate && pip install cf-units
New python executable in /tmp/venv/bin/python
Installing setuptools, pip, wheel...done.
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting cf-units
  Downloading https://files.pythonhosted.org/packages/43/45/c23e9412fbccf511de996bb6450ca8862e1273839617dd17240d9c179324/cf-units-2.1.1.tar.gz (174kB)
     |████████████████████████████████| 184kB 3.1MB/s 
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-COYfrD/cf-units/setup.py", line 144, in <module>
        ext_modules=[udunits_ext]
      File "/tmp/venv/local/lib/python2.7/site-packages/setuptools/__init__.py", line 144, in setup
        _install_setup_requires(attrs)
      File "/tmp/venv/local/lib/python2.7/site-packages/setuptools/__init__.py", line 139, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "/tmp/venv/local/lib/python2.7/site-packages/setuptools/dist.py", line 717, in fetch_build_eggs
        replace_conflicting=True,
      File "/tmp/venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 782, in resolve
        replace_conflicting=replace_conflicting
[...]

Issue #151, opened 2019-05-07
AttributeError: 'dict' object has no attribute '__NUMPY_SETUP__'


ghost:
Since https://github.com/SciTools/cf-units/pull/145 was merged (which I was happy to note, it was pretty annoying having to pre-install Python dependencies!), I get this error when attempting to run unit tests on a package which depends on cf-units.

It looks to be a relatively common issue with the `__builtins__` object sometimes being a module, and sometimes being a dict, and any code accessing it must handle *both* of these cases.

Other examples:
https://github.com/pycalphad/pycalphad/issues/26
https://github.com/erikrose/nose-progressive/pull/54

I've included this very contrived example to demonstrate the problem:

```bash
$ mkdir /tmp/mypackage && cd /tmp/mypackage

$ virtualenv venv && source venv/bin/activate
New python executable in /tmp/mypackage/venv/bin/python
Installing setuptools, pip, wheel...done.

$ cat setup.py
from setuptools import setup
setup(
    name='mypackage',
    version='0.1',
    tests_require='cf-units'
)

$ python setup.py test
running test
Searching for cf-units
Reading https://pypi.org/simple/cf-units/
Downloading https://files.pythonhosted.org/packages/43/45/c23e9412fbccf511de996bb6450ca8862e1273839617dd17240d9c179324/cf-units-2.1.1.tar.gz#sha256=fa0ef8efd84546e61088aa23e76ebbaf7043167dc3a7f35f34549c234b543530
Best match: cf-units 2.1.1
Processing cf-units-2.1.1.tar.gz
Writing /tmp/easy_install-Smrlyk/cf-units-2.1.1/setup.cfg
Running cf-units-2.1.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-Smrlyk/cf-units-2.1.1/egg-dist-tmp-mywrU6
Traceback (most recent call last):
  File "setup.py", line 6, in <module>
    tests_require='cf-units'
  File "/tmp/mypackage/venv/local/lib/python2.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
[...]

PR #153, opened 2019-07-22
Numpy setup.py fixes

ghost:
I've provided some examples of how I was able to fix the couple of issues (#151 and #152) I recently raised in this repo.

Feel free to pick them apart or use as-is, or suggest changes!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,8 +98,15 @@ def numpy_build_ext(pars):
 
     class build_ext(_build_ext):
         def finalize_options(self):
+            # See https://github.com/erikrose/nose-progressive/pull/54
+            def _set_builtin(name, value):
+                if isinstance(__builtins__, dict):
+                    __builtins__[name] = value
+                else:
+                    setattr(__builtins__, name, value)
+
             _build_ext.finalize_options(self)
-            __builtins__.__NUMPY_SETUP__ = False
+            _set_builtin('__NUMPY_SETUP__', False)
             import numpy
             self.include_dirs.append(numpy.get_include())
 
--- a/setup.py
+++ b/setup.py
@@ -144,7 +144,7 @@ setup(
     package_data={'cf_units': list(file_walk_relative('cf_units/etc',
                                                       remove='cf_units/'))},
     install_requires=load('requirements.txt'),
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -99,7 +106,7 @@ def numpy_build_ext(pars):
     class build_ext(_build_ext):
         def finalize_options(self):
             _build_ext.finalize_options(self)
-            __builtins__.__NUMPY_SETUP__ = False
+            set_builtin('__NUMPY_SETUP__', False)
```
bjlittle: @lwgordonimos Would you mind making this a local function to the `build_ext.finalize_options` method? i.e.
```python
class build_ext(_build_ext):
    def finalize_options(self):
        # See https://github.com/SciTools/cf-units/issues/151
        def _set_builtin(name, value):
            if isinstance(__builtins__, dict):
                __builtins__[name] = value
            else:
                setattr(__builtins__, name, value)

        _build_ext.finalize_options(self)
        _set_builtin('__NUMPY_SETUP__', False)
        import numpy
        self.include_dirs.append(numpy.get_include())
```
as it's only really relevant to the `build_ext.finalize_options` method, thanks.

ghost: Not at all, done!

bjlittle: @lwgordonimos Nice one, thanks... but you've made the `set_builtin` function local to the `numpy_build_ext` factory function, rather than the `finalize_options` method in the `build_ext` class, see my example above :wink: 

I'm keen to keep the `set_builtin` function tightly coupled to where it is used i.e. `build_ext.finalize_options`

Would you be happy to do that? Much obliged :smile:

bjlittle:
@lwgordonimos Awesome, thanks :+1:

ghost:
No worries, makes perfect sense to keep it contained. Sorry, I glanced at your example and jumped the gun with copy/pasting without noting the details!

bjlittle:
@lwgordonimos Cool, thanks for the PR, much appreciated! :beers: 

I've already tagged the patch release `2.1.2`, and I'm in the process of building and uploading to PyPI, and I'll also update the recipe on `conda-forge`, so hopefully this will unblock you.

Nice one, all the best!

PR #156, opened 2020-02-18
remove cython as a run dependency

ocefpaf:
The current setup requires `cython` to be install at run time and  that breaks the `pip check` and other dependency checking commands.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,5 @@
 coveralls
+cython
 jinja2
 pep8
 pip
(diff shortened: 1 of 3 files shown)


ocefpaf:
Just would like to ping this again for a review. This came up today again rendering some development envs broken b/c people did not have cython in the, even though though don't need it.

ocefpaf:
Ping again... @bjlittle having this merged and a new release would really help our CIs that use this package.

ajdawson:
Thanks for sorting this out @ocefpaf

ocefpaf:
Thanks @ajdawson! I'll prepare a micro bugfix release if you all are OK with it.

PR #164, opened 2021-03-01
Type preservation

### LINKED ISSUE #163: Casting to integer on conversion
Hi there,

I have a question about behaviour which I was hoping you could help with. It focuses on this comment:

https://github.com/SciTools/cf-units/blob/a3eeaf15622bd3ec6f23c18cbd3b3658086aa7cf/cf_units/__init__.py#L1845

My question is: why isn't this also applied to `np.float64` datatypes? 

More info:

The minimal example below shows how this can cause a `np.float64` dtype to be cast to integer even though a `np.float32` dtype is not. This seems surprising, but maybe that is the idea? (As a sidenote, this only appears to happen on a linux operating system, it doesn't happen on mac...)

```python
In [1]: import cf_units

In [2]: import numpy as np

In [3]: cf_units.__version__
Out[3]: '2.1.4'

In [4]: start = cf_units.Unit('days since 2100-01-01', calendar='365_day')

In [5]: target = cf_units.Unit('days since 2006-01-01', calendar='365_day')

In [6]: start.convert(np.array([1, 2, 3]).astype(np.float32), target)
Out[6]: array([34311., 34312., 34313.], dtype=float32)

In [7]: start.convert(np.array([1, 2, 3]).astype(np.float32), target).dtype
Out[7]: dtype('float32')  # type preserved

In [8]: start.convert(np.array([1, 2, 3]).astype(np.float64), target)
Out[8]: array([34311, 34312, 34313])

In [9]: start.convert(np.array([1, 2, 3]).astype(np.float64), target).dtype
Out[9]: dtype('int64')  # type not preserved
```

Thanks!

znicholls:
Closes #163

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1738,7 +1738,11 @@ class Unit(_OrderedHashable):
                     result, self.cftime_unit, self.calendar)
                 result = cftime.date2num(
                     result_datetimes, other.cftime_unit, other.calendar)
-                if isinstance(value, np.ndarray) and np.issubsctype(value.dtype, np.floating):
+                convert_type = (
+                    isinstance(value, np.ndarray)
+                    and np.issubsctype(value.dtype, np.floating)
+                )
+                if convert_type:
                     result = result.astype(value.dtype)
             else:
                 try:
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1738,10 +1738,8 @@ class Unit(_OrderedHashable):
                     result, self.cftime_unit, self.calendar)
                 result = cftime.date2num(
                     result_datetimes, other.cftime_unit, other.calendar)
-                # Preserve the datatype of the input array if it was float32.
-                if (isinstance(value, np.ndarray) and
-                   value.dtype == np.float32):
-                    result = result.astype(np.float32)
+                if isinstance(value, np.ndarray) and np.issubsctype(value.dtype, np.floating):
+                    result = result.astype(value.dtype)
             else:
                 try:
                     ut_converter = _ud.get_converter(self.ut_unit,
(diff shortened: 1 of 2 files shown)


cf_units/__init__.py
```diff
@@ -1842,10 +1842,12 @@ def convert(self, value, other, ctype=FLOAT64, inplace=False):
...
-                # Preserve the datatype of the input array if it was float32.
-                if (isinstance(value, np.ndarray) and
-                   value.dtype == np.float32):
-                    result = result.astype(np.float32)
+                # Preserve the datatype of the input array if it was float.
+                if isinstance(value, np.ndarray):
+                    if value.dtype == np.float32:
+                        result = result.astype(np.float32)
+                    elif value.dtype == np.float64:
+                        result = result.astype(np.float64)
```
ocefpaf: @bjlittle are you OK with these changes? I have to confess I'm not sure why that was there in the first place. Maybe we should just always convert to the original dtype?

rcomer: Here is the original addition: https://github.com/SciTools/iris/pull/1385

So it was about making the time coordinate consistent for all cubes in a cubelist.  So it seems to me that we should also preserve float64 (but I have no authority here!)

I think trying to preserve ints could be problematic: if you convert from say “hours since ...” to “days since...” and not everything is midnight, the output would need to be float.

trexfeathers: Thanks for your wisdom @rcomer, I agree. Most of the original cf-units devs have moved on so we all have about as much authority as each other here!

rcomer: @trexfeathers when I wrote that comment I hadn't yet contributed in `cf-units`, so I probably have more authority now than I did then!

trexfeathers:
```suggestion
                if isinstance(value, np.ndarray) and np.issubsctype(value, float):
                    result = result.astype(value.dtype)
```

trexfeathers (changes requested):
Thanks so much for your contribution @znicholls! And sorry for letting this go stale.

Please could you rebase, and service my suggestion for making the code more generic?

cf_units/__init__.py
znicholls: Nice done

znicholls:
> sorry for letting this go stale

No worries, lots happening

> Please could you rebase, and service my suggestion for making the code more generic?

Done

trexfeathers (approved):

PR #165, opened 2021-04-22
num2date: return datetimes by default if possible

### LINKED ISSUE #161: Change of behaviour due to latest cftime
It seems that the latest version of `cftime` always returns cftime objects, which means that the cf-units documentation is no longer correct and the default behaviour has changed.

Specifically, the `only_use_cftime_datetimes` option in `cftime.num2date` defaults to `True`. I would argue that the cf-units call to this function should set this to be `False` for the meantime to maintain the current documented expectation that it should return python datetimes where possible. In a future major release it would be better to allow this option (and the accompanying `only_use_python_datetimes` option) to be passed down the call stack.

rcomer:
Aims to fix #161.  I don't know how to run the tests (or anything else) locally with my branch, so I am trusting to Travis!  I did run `flake8` locally on the files I’ve changed, and it only complained about two lines that I haven’t touched.


rcomer:
There is Test Carnage, but I'm not (yet) convinced it's because I've done something wrong...

cf_units/__init__.py
```diff
@@ -607,8 +623,11 @@ def _num2date_to_nearest_second(time_value, utime):
     ceil_mask = np.logical_or(has_half_seconds, microseconds >= 500000)
     if time_values[ceil_mask].size > 0:
         useconds = Unit('second')
-        second_frac = useconds.convert(0.75, utime.units)
```
rcomer, line 605: I don't know what to do about this line.  Need a replacement for `utime.units`.

rcomer:
Right.  The reason my original effort didn't work was because we call `cftime.utime.num2date`, which doesn't appear to accept the `only_use_cftime_datetimes` keyword.  So I have switched to using the `cftime.num2date` function.

To do this properly, we should probably sort out #166, but I think that's beyond the scope of this PR.

rcomer:
Why are tests not running for my latest commit? :confused:

rcomer:
All tests now passing, except for some doctests (and a whole raft of python2.7 tests) which also fail at #164, so not caused by this change.

rcomer:
Rebased.  Tests now all passing 🙂

trexfeathers:
Thanks for looking into this @rcomer 💐

Having discussed with a few devs, **I don't want to merge this to `master`**:

To avoid future difficulties/irrelevance a package should evolve with its dependencies, so as @duncanwp hinted at it in #161: the next major release of cf-units should explicitly adopt the new default behaviour of `cftime.num2date`. That's what should be going into `master`. It would certainly be great to provide the `only_use_cftime_datetimes` parameter, but again the default value should align with ctime.

Pinning the behaviour to how it was in the 'old world' is a **bugfix**, enabling continued use of the existing major release of cf-units. As I understand it, bugfix changes can/should only be applied to a bugfix branch, and that certainly seems appropriate here given the conflict with an opposite intended future behaviour.

I'm just making sure I fully understand how this ought to be done. I had started working on a migration to Cirrus+Nox, but at least for now Travis seems to be playing nicely so I'll only involve that if I need to!

rcomer:
Thanks @trexfeathers, I shall await further instructions.

trexfeathers (changes requested):
In terms of the code, this looks good. One change if you can:

Also need an equivalent change when using `cftime.utime` - [L1937](https://github.com/SciTools/cf-units/pull/165/files#diff-ef037721b197e608853e2e25efd260fb32471e5e861083e0c2996fecbde976d6R1937), and propagate upwards for the container method and uses.

You're definitely right to move away from using `cftime.utime` where possible, since it's on its way out. But as of `v1.4.1` it's still in the cftime API and it also has a `only_use_cftime_datetimes` parameter - [see here](https://github.com/Unidata/cftime/blob/204a23451dd1bd97205e7e9f2da958eccaea5a12/src/cftime/_cftime_legacy.pyx#L446-L448).

rcomer:
@trexfeathers have I understood correctly?  If the `utime` object contains the `only_use_cftime_datetimes` flag then we don't need the changes to `_num2date_to_nearest_second` or associated test changes.

Unfortunately, having made and reverted changes to `test__num2date_to_nearest_second.py`, I have confused the license header checker.

trexfeathers:
>@trexfeathers have I understood correctly?  If the `utime` object contains the `only_use_cftime_datetimes` flag then we don't need the changes to `_num2date_to_nearest_second` or associated test changes.

Oh dear, I was only suggesting an additional change, not undoing your existing! While you are correct, I was trying to say that your changes up to that point had been valuable anyway:

>You're definitely right to move away from using `cftime.utime` where possible, since it's on its way out.

rcomer:
Okedoke.  Commit histories are fun!

I believe [this](https://github.com/SciTools/cf-units/pull/165/files/e54c8cdffe2d8f89fc78537f78a80a20517f36fd..bf6a0de82aaa49a643d52f53e0415e722cb92e35) shows the changes since your initial review.

Does the `utime` method need a test?  I can't find any existing ones for that method...

rcomer:
Also, do you have a "squash" option in your merge button or should I do that manually?

trexfeathers (approved):
Looks good @rcomer. If you can target the new [`v2.1.5` branch](https://github.com/SciTools/cf-units/tree/v2.1.5) I'm good to merge.

>Does the `utime` method need a test?  I can't find any existing ones for that method...

Since you didn't introduce this method, and given its future limited life in cftime, I'm happy to forego additional tests.

>Also, do you have a "squash" option in your merge button or should I do that manually?

I do have the squash option - I just need to remember to use it!

rcomer:
Thanks @trexfeathers, now re-targetted.

PR #157, opened 2020-02-18
drop python 2

ocefpaf:
This should simplify our deployment of things that depend on `cf-units`.

PR #167, opened 2021-04-23
Fix doctests

rcomer:
Some tweaks to get the doctests passing.  Added in 30 minutes to the datetimes that are input to `date2num` so we'll get a decimal and therefore more obviously a float in the output.  Also removed the `round` from the Unit method version so it's consistent with the function.

For failure example see https://travis-ci.org/github/SciTools/cf-units/jobs/768092165#L1103

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -442,14 +442,14 @@ def date2num(date, unit, calendar):
 
         >>> import cf_units
         >>> import datetime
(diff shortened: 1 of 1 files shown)


rcomer:
Travis just not running.

rcomer:
Python 3 tests now all passing.  Python 2 tests are a mess, but we have #157 to sort that out 👍.

rcomer:
I have a full set of green ticks 🙂

trexfeathers (approved):

PR #168, opened 2021-05-25
Updates for cftime 1.5.0

### LINKED ISSUE #159: Incompatibility of num2date
The [description](https://github.com/SciTools/cf-units/blob/a3eeaf15622bd3ec6f23c18cbd3b3658086aa7cf/cf_units/__init__.py#L511) of `cf_units.num2date()` claims that it behaves the same as `matplotlib.dates.num2date()`, which [is](https://github.com/matplotlib/matplotlib/blob/afa8d2821d0189b809ce3fd4ed4beabe5c5f97a8/lib/matplotlib/dates.py#L436) [itself](https://github.com/matplotlib/matplotlib/blob/afa8d2821d0189b809ce3fd4ed4beabe5c5f97a8/lib/matplotlib/dates.py#L304) [a call](https://github.com/matplotlib/matplotlib/blob/afa8d2821d0189b809ce3fd4ed4beabe5c5f97a8/lib/matplotlib/dates.py#L283) to `datetime.datetime.fromordinal()`.

Comparing these calls for day 1,
```python
from cf_units import Unit
from matplotlib.dates import num2date
from datetime import datetime

pg_unit = Unit('days since 001-01-01 00:00:00', calendar='proleptic_gregorian')
print("cfunits.num2date():", pg_unit.num2date(1))
print("matplotlib.dates.num2date():", num2date(1))
print("datetime.datetime.fromordinal():", datetime.fromordinal(1))
```

> cfunits.num2date(): 0001-01-02 00:00:00
> matplotlib.dates.num2date(): 0001-01-01 00:00:00+00:00
> datetime.datetime.fromordinal(): 0001-01-01 00:00:00

The easy answer would be to change the documentation, but I feel like naive use of `num2date()` should replicate the `datetime` behaviour.

### LINKED ISSUE #161: Change of behaviour due to latest cftime

### LINKED ISSUE #166: Stop using cftime.utime
The `Unit.utime()` method returns a `ctime.utime` object, which is then used for all the `cf_units` time conversions.  This object has been removed as "legacy" for the next `cftime` release: https://github.com/Unidata/cftime/pull/235

rcomer:
Closes #166.  As well as the `utime` object, the julian date functions were also removed at https://github.com/Unidata/cftime/pull/235 and the note in the [changelog](https://github.com/Unidata/cftime/blob/master/Changelog) says they are replaced by `toordinal` and `fromordinal`.

Note that the first commit here is from #165, which was included in v2.1.5.  I have then changed the default behaviour of the various `num2date` functions to return a `cftime` object, consistent with the `cftime.num2date` function.  This goes some way to closing #161, though I have not here exposed the `only_use_python_datetimes` option.  I agree that that would be desirable, but it would be easy to add in a separate PR and I think there is already more than enough here for the reviewer to wade through!

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -506,8 +506,9 @@ def num2date(time_value, unit, calendar, only_use_cftime_datetimes=True):
 
     Like the :func:`matplotlib.dates.num2date` function, except that it allows
     for different units and calendars.  If
-    unit = 'days since 1970-01-01 00:00:00' and
-    calendar = 'proleptic_gregorian',
+    unit = 'days since 1970-01-01 00:00:00',
+    calendar = 'proleptic_gregorian' and
+    only_use_cftime_datetimes = False
     behaves the same as matplotlib with the default epoch of 1970-01-01.
 
     By default, the datetime instances returned are cftime.datetime objects,
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -569,7 +569,7 @@ def num2date(time_value, unit, calendar, only_use_cftime_datetimes=True):
         time_value, only_use_cftime_datetimes=only_use_cftime_datetimes)
 
 
-def _num2date_to_nearest_second(time_value, utime,
+def _num2date_to_nearest_second(time_value, unit,
                                 only_use_cftime_datetimes=True):
     """
     Return datetime encoding of numeric time value with respect to the given
@@ -577,8 +577,8 @@ def _num2date_to_nearest_second(time_value, utime,
 
     * time_value (float):
         Numeric time value/s.
-    * utime (cftime.utime):
-        cftime.utime object with which to perform the conversion/s.
+    * unit (Unit):
+        cf_units.Unit object with which to perform the conversion/s.
 
     * only_use_cftime_datetimes (bool):
         If True, will always return cftime datetime objects, regardless of
@@ -587,6 +587,7 @@ def _num2date_to_nearest_second(time_value, utime,
 
     Returns:
         datetime, or numpy.ndarray of datetime object.
+        
     """
     time_values = np.asanyarray(time_value)
     shape = time_values.shape
@@ -601,10 +602,12 @@ def _num2date_to_nearest_second(time_value, utime,
     # those versions, a half-second value would be rounded up or down
     # arbitrarily. It is probably not possible to replicate that behaviour with
     # later versions, if one wished to do so for the sake of consistency.
-    has_half_seconds = np.logical_and(utime.units == 'seconds',
+    cftime_unit = unit.cftime_unit
+    time_units = cftime_unit.split(' ')[0]
+    has_half_seconds = np.logical_and(time_units == 'seconds',
                                       time_values % 1. == 0.5)
     dates = cftime.num2date(
-        time_values, utime.unit_string, calendar=utime.calendar,
+        time_values, cftime_unit, calendar=unit.calendar,
         only_use_cftime_datetimes=only_use_cftime_datetimes)
     try:
         # We can assume all or none of the dates have a microsecond attribute
@@ -615,10 +618,10 @@ def _num2date_to_nearest_second(time_value, utime,
     ceil_mask = np.logical_or(has_half_seconds, microseconds >= 500000)
     if time_values[ceil_mask].size > 0:
         useconds = Unit('second')
-        second_frac = useconds.convert(0.75, utime.units)
+        second_frac = useconds.convert(0.75, time_units)
         dates[ceil_mask] = cftime.num2date(
-            time_values[ceil_mask] + second_frac, utime.unit_string,
-            calendar=utime.calendar,
+            time_values[ceil_mask] + second_frac, cftime_unit,
+            calendar=unit.calendar,
             only_use_cftime_datetimes=only_use_cftime_datetimes)
     dates[round_mask] = _discard_microsecond(dates[round_mask])
     result = dates[0] if shape is () else dates.reshape(shape)
@@ -1823,13 +1826,14 @@ class Unit(_OrderedHashable):
                 result = value
             else:
                 result = copy.deepcopy(value)
-            # Use utime for converting reference times that are not using a
+            # Use cftime for converting reference times that are not using a
             # gregorian calendar as it handles these and udunits does not.
             if self.is_time_reference() \
                     and self.calendar != CALENDAR_GREGORIAN:
-                ut1 = self.utime()
-                ut2 = other.utime()
-                result = ut2.date2num(ut1.num2date(result))
+                result_datetimes = cftime.num2date(
+                    result, self.cftime_unit, self.calendar)
+                result = cftime.date2num(
+                    result_datetimes, other.cftime_unit, other.calendar)
                 # Preserve the datatype of the input array if it was float32.
                 if (isinstance(value, np.ndarray) and
                    value.dtype == np.float32):
@@ -1894,34 +1898,11 @@ class Unit(_OrderedHashable):
             raise ValueError("Unable to convert from '%r' to '%r'." %
                              (self, other))
 
-    def utime(self, only_use_cftime_datetimes=False):
+    @property
+    def cftime_unit(self):
         """
-        Returns a cftime.utime object which performs conversions of
-        numeric time values to/from datetime objects given the current
-        calendar and unit time reference.
-
-        The current unit time reference must be of the form:
-        '<time-unit> since <time-origin>'
-        i.e. 'hours since 1970-01-01 00:00:00'
-
-        Kwargs:
-
-        * only_use_cftime_datetimes (bool):
-            If True, num2date method will always return cftime datetime
-            objects, regardless of calendar.  If False, returns
-            datetime.datetime instances where possible.  Defaults to False.
-
-        Returns:
-            cftime.utime.
-
-        For example:
-
-            >>> import cf_units
-            >>> u = cf_units.Unit('hours since 1970-01-01 00:00:00',
-            ...                   calendar=cf_units.CALENDAR_STANDARD)
-            >>> ut = u.utime()
-            >>> print(ut.num2date(2))
-            1970-01-01 02:00:00
+        Returns a string suitable for passing as a unit to cftime.num2date and
+        cftime.date2num.
 
         """
         if self.calendar is None:
@@ -1938,9 +1919,7 @@ class Unit(_OrderedHashable):
         # ensure to strip out non-parsable 'UTC' postfix, which
         # is generated by UDUNITS-2 formatted output
         #
-        return cftime.utime(
-            str(self).rstrip(" UTC"), self.calendar,
-            only_use_cftime_datetimes=only_use_cftime_datetimes)
+        return str(self).rstrip(" UTC")
 
     def date2num(self, date):
         """
@@ -1977,9 +1956,8 @@ class Unit(_OrderedHashable):
 
         """
 
-        cdf_utime = self.utime()
         date = _discard_microsecond(date)
-        return cdf_utime.date2num(date)
+        return cftime.date2num(date, self.cftime_unit, self.calendar)
 
     def num2date(self, time_value, only_use_cftime_datetimes=True):
         """
@@ -2028,9 +2006,7 @@ class Unit(_OrderedHashable):
             ['1970-01-01 06:00:00', '1970-01-01 07:00:00']
 
         """
-        cdf_utime = self.utime(
-            only_use_cftime_datetimes=only_use_cftime_datetimes)
 
         return _num2date_to_nearest_second(
-            time_value, cdf_utime,
+            time_value, self,
             only_use_cftime_datetimes=only_use_cftime_datetimes)
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -587,7 +587,7 @@ def _num2date_to_nearest_second(time_value, unit,
 
     Returns:
         datetime, or numpy.ndarray of datetime object.
-        
+
     """
     time_values = np.asanyarray(time_value)
     shape = time_values.shape
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -484,7 +484,7 @@ def _discard_microsecond(date):
     dates = np.asarray(date)
     shape = dates.shape
     dates = dates.ravel()
-    # Create date objects of the same type returned by utime.num2date()
+    # Create date objects of the same type returned by cftime.num2date()
     # (either datetime.datetime or cftime.datetime), discarding the
     # microseconds
     dates = np.array([d and d.__class__(d.year, d.month, d.day,
@@ -494,7 +494,7 @@ def _discard_microsecond(date):
     return result
 
 
-def num2date(time_value, unit, calendar):
+def num2date(time_value, unit, calendar, only_use_cftime_datetimes=False):
     """
     Return datetime encoding of numeric time value (resolution of 1 second).
 
@@ -508,7 +508,7 @@ def num2date(time_value, unit, calendar):
     unit = 'days since 001-01-01 00:00:00'}
     calendar = 'proleptic_gregorian'.
 
-    The datetime instances returned are 'real' python datetime
+    By default, the datetime instances returned are 'real' python datetime
     objects if the date falls in the Gregorian calendar (i.e.
     calendar='proleptic_gregorian', or calendar = 'standard' or 'gregorian'
     and the date is after 1582-10-15). Otherwise, they are 'phony' datetime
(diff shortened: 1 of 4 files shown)


cf_units/__init__.py
```diff
@@ -398,13 +398,13 @@ def date2julian_day(date, calendar):
...
         >>> import datetime
-        >>> cf_units.date2julian_day(datetime.datetime(1970, 1, 1, 0, 0, 0),
+        >>> cf_units.date2julian_day(cftime.datetime(1970, 1, 1, 0, 0, 0),
         ...                          cf_units.CALENDAR_STANDARD)
         2440587.5...
 
     """
 
-    return cftime.JulianDayFromDate(date, calendar)
+    return date.toordinal(fractional=True)
```
rcomer: Note the use of the `fractional` keyword here means that the function will no longer work if a python `datetime.datetime` object is passed.  I note that the docstring specifies a `cftime.date` object should be passed, but the doctest previously used a `datetime.datetime`.  Without the `fractional` keyword, the doctest still failed on the `datetime.datetime`, because a very different number was returned.  Open to suggestions on the best thing to do here!

cf_units/__init__.py
```diff
@@ -359,14 +359,14 @@ def julian_day2date(julian_day, calendar):
...
-        >>> dt = datetime.datetime(1970, 1, 1, 0, 0, 0)
+        >>> dt = cftime.datetime(1970, 1, 1, 0, 0, 0)
         >>> jd = cf_units.date2julian_day(dt, cf_units.CALENDAR_STANDARD)
         >>> print(cf_units.julian_day2date(jd, cf_units.CALENDAR_STANDARD))
         1970-01-01 00:00:00
 
     """
 
-    return cftime.DateFromJulianDay(julian_day, calendar)
+    return cftime.datetime.fromordinal(julian_day, calendar)
```
rcomer: Have just realised that this docstring is wrong, and possibly has been for some time, as "real" `datetime.datetime` instances are not returned even with `cftime.DateFromJulianDay` under cftime v1.2.1.

Given that these julian day functions just pass their args straight to the relevant `cftime` functions, I'm wondering if there is any real benefit in maintaining them and trying to keep the docstrings up-to-date with `cftime` changes.  Could we just remove them and direct users to use the `cftime` functions directly instead?

cf_units/__init__.py
```diff
@@ -508,14 +508,11 @@ def num2date(time_value, unit, calendar):
     unit = 'days since 001-01-01 00:00:00'}
     calendar = 'proleptic_gregorian'.
```
trexfeathers: The beginning of this paragraph (which I can't highlight!) is inaccurate since we're actually wrapping/mimicking `cftime.num2date`. If you can get this lined up we'll close #159 in the process 👍

rcomer: Thanks @trexfeathers, looks like the `maplotlib` behaviour has changed since #159 was opened.  How's this?

cf_units/__init__.py
```diff
@@ -494,28 +495,26 @@ def _discard_microsecond(date):
...
+def num2date(time_value, unit, calendar, only_use_cftime_datetimes=True):
     """
     Return datetime encoding of numeric time value (resolution of 1 second).
 
     The units of the numeric time value are described by the unit and
     calendar arguments. The returned datetime object represent UTC with
     no time-zone offset, even if the specified unit contain a time-zone
     offset.
-
+
```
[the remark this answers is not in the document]
rcomer: I have become too used to Iris's pre-commit hooks!

cf_units/__init__.py
rcomer: mpl ref for info: https://matplotlib.org/stable/api/dates_api.html#matplotlib-date-format

cf_units/__init__.py
```diff
@@ -504,18 +505,17 @@ def num2date(time_value, unit, calendar):
...
-    and the date is after 1582-10-15). Otherwise, they are 'phony' datetime
-    objects which support some but not all the methods of 'real' python
-    datetime objects.  This is because the python datetime module cannot
-    use the 'proleptic_gregorian' calendar, even before the switch
-    occured from the Julian calendar in 1582. The datetime instances
+    for different units and calendars.  If
+    unit = 'days since 1970-01-01 00:00:00',
+    calendar = 'proleptic_gregorian' and
+    only_use_cftime_datetimes = False
+    behaves the same as matplotlib with the default epoch of 1970-01-01.
```
trexfeathers: Despite all these specifics, the behaviour is still not analogous (see below).

Given this, and the fact that the version in `cf_units` has several extra parameters, there seems little point in making the claim about being like `matplotlib.dates.num2date`. How would you feel about removing the paragraph?

```python
from matplotlib.dates import num2date as num2date_mpl
from cf_units import num2date as num2date_cfu

num2date_cfu([1., 2.], unit='days since 1970-01-01 00:00:00', calendar='proleptic_gregorian', only_use_cftime_datetimes=False)
# array([datetime.datetime(1970, 1, 2, 0, 0), datetime.datetime(1970, 1, 3, 0, 0)], dtype=object)

num2date_mpl([1., 2.])
# [datetime.datetime(1970, 1, 2, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(1970, 1, 3, 0, 0, tzinfo=datetime.timezone.utc)]

```

cf_units/__init__.py
trexfeathers: **Argument for keeping:** cf-units hasn't always used cftime underneath (#85), and we can't guarantee it always will. These functions can protect the user from API changes.

**Argument for removal:** cftime is moving much faster than cf-units and it will be continual effort to keep these functions aligned.

For me the removal argument is much stronger. That's a 2-to-0 vote, go for it!

cf_units/__init__.py
trexfeathers: Let's remove the function altogether!

cf_units/__init__.py
```diff
@@ -418,9 +418,10 @@ def date2num(date, unit, calendar):
     applied to the returned numeric values.
 
     Like the :func:`matplotlib.dates.date2num` function, except that it allows
-    for different units and calendars.  Behaves the same as if
-    unit = 'days since 0001-01-01 00:00:00' and
-    calendar = 'proleptic_gregorian'.
+    for different units and calendars.  If
+    unit = 'days since 1970-01-01 00:00:00' and
+    calendar = 'proleptic_gregorian',
+    behaves the same as matplotlib with the default epoch of 1970-01-01.
```
trexfeathers: See the `num2date` comment.

cf_units/__init__.py
rcomer: Except it's still not _quite_ the same since `cftime.num2date` returns a subclass of `datetime.datetime`.  Also we only work at second precision whereas `matplotlib` works at microsecond precision.

We could just remove this paragraph entirely?  Any user who is concerned about how our behaviour compares to `matplotlib` behaviour can always do their own comparison.

cf_units/__init__.py
```diff
@@ -561,38 +568,50 @@ def num2date(time_value, unit, calendar):
...
     # later versions, if one wished to do so for the sake of consistency.
-    has_half_seconds = np.logical_and(utime.units == 'seconds',
+    cftime_unit = unit.cftime_unit
+    time_units = cftime_unit.split(' ')[0]
+    has_half_seconds = np.logical_and(time_units == 'seconds',
                                       time_values % 1. == 0.5)
-    dates = utime.num2date(time_values)
+    dates = cftime.num2date(
+        time_values, cftime_unit, calendar=unit.calendar,
+        only_use_cftime_datetimes=only_use_cftime_datetimes)
```
trexfeathers: Store `units`, `calendar` and `only_use_cftime_datetimes` in a `dict` - it's presumably important that both uses of `num2date` in this function are configured identically.

cf_units/tests/integration/test__num2date_to_nearest_second.py
```diff
@@ -101,39 +105,60 @@ def test_simple_gregorian(self):
...
+                 self.uhours, self.uhours,
+                 self.udays, self.udays]
+        expected = [cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 20),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 40),
+                    cftime.DatetimeGregorian(1970, 1, 1, 1, 15),
+                    cftime.DatetimeGregorian(1970, 1, 1, 2, 30),
+                    cftime.DatetimeGregorian(1970, 1, 1, 8),
+                    cftime.DatetimeGregorian(1970, 1, 1, 16),
+                    cftime.DatetimeGregorian(1970, 10, 28),
+                    cftime.DatetimeGregorian(1971, 8, 24)]
```
trexfeathers: [Specific calendar classes have been deprecated](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/__init__.py#L8-L11). Anything new we introduce should therefore be `cftime`'s standard `datetime` object with an appropriate `calendar` parameter.

cf_units/tests/integration/test__num2date_to_nearest_second.py
```diff
@@ -101,39 +105,60 @@ def test_simple_gregorian(self):
...
-        self.check_dates(nums, utimes, expected)
+        units = [self.uminutes, self.uminutes,
+                 self.uhours, self.uhours,
+                 self.udays, self.udays]
+        expected = [cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 5),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 10),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 15),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 30),
+                    cftime.DatetimeGregorian(1970, 1, 1, 8),
+                    cftime.DatetimeGregorian(1970, 1, 1, 16)]
```
trexfeathers: [Specific calendar classes have been deprecated](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/__init__.py#L8-L11). Anything new we introduce should therefore be `cftime`'s standard `datetime` object with an appropriate `calendar` parameter.

cf_units/tests/integration/test__num2date_to_nearest_second.py
```diff
@@ -101,39 +105,60 @@ def test_simple_gregorian(self):
...
-                    datetime.datetime(1970, 1, 1, 0, 0, 4),
-                    datetime.datetime(1970, 1, 1, 0, 0, 5)]
+        units = [self.useconds] * 7
+        expected = [cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 0),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 1),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 1),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 2),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 3),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 4),
+                    cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 5)]
```
trexfeathers: [Specific calendar classes have been deprecated](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/__init__.py#L8-L11). Anything new we introduce should therefore be `cftime`'s standard `datetime` object with an appropriate `calendar` parameter.

cf_units/tests/test_unit.py
```diff
@@ -986,7 +987,16 @@ class TestNumsAndDates(unittest.TestCase):
     def test_num2date(self):
         u = Unit('hours since 2010-11-02 12:00:00',
                  calendar=unit.CALENDAR_STANDARD)
-        self.assertEqual(str(u.num2date(1)), '2010-11-02 13:00:00')
+        res = u.num2date(1)
+        self.assertEqual(str(res), '2010-11-02 13:00:00')
+        self.assertIsInstance(res, cftime.DatetimeGregorian)
```
trexfeathers: [Specific calendar classes have been deprecated](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/__init__.py#L8-L11). Anything new we introduce should therefore be `cftime`'s standard `datetime` object with an appropriate `calendar` parameter.

cf_units/tests/integration/test__num2date_to_nearest_second.py
```diff
@@ -22,76 +22,80 @@
...
+        self.uminutes = cf_units.Unit('minutes since 1970-01-01', calendar)
+        self.uhours = cf_units.Unit('hours since 1970-01-01', calendar)
+        self.udays = cf_units.Unit('days since 1970-01-01', calendar)
+
+    def check_dates(self, nums, units, expected, only_cftime=True):
+        for num, unit, exp in zip(nums, units, expected):
+            res = _num2date_to_nearest_second(
+                num, unit, only_use_cftime_datetimes=only_cftime)
             self.assertEqual(exp, res)
+            self.assertIsInstance(res, type(exp))
```
trexfeathers, line 41: Given my comments below about the [deprecation of specific calendar classes](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/__init__.py#L8-L11), we may need to switch to interrogating the `calendar` attribute of `res` instead of looking at `type`.

trexfeathers (changes requested):
Thanks so much @rcomer, this is great work, and I'm very glad that someone as experienced as you with units has been working on this problem!

A few changes if you don't mind.

cf_units/__init__.py
trexfeathers: https://github.com/SciTools/cf-units/pull/168#discussion_r647389483

cf_units/tests/integration/test__num2date_to_nearest_second.py
rcomer: Ah, nuts!

cf_units/tests/integration/test__num2date_to_nearest_second.py
rcomer: Looks like equality assertion will fail if calendars are different:
```python
In [16]: cftime.datetime(1970, 1, 2, calendar='gregorian') == cftime.datetime(1970, 1, 2, calendar='360_day')
Out[16]: False
```

trexfeathers: So ironically the original implementation was accidentally the right one!

rcomer: But
```python
In [5]: datetime.datetime(1970, 1, 2) == cftime.datetime(1970, 1, 2, calendar='gregorian')
Out[5]: True
```
So I think the type check still makes sense.

cf_units/__init__.py
```diff
@@ -602,10 +524,11 @@ def _num2date_to_nearest_second(time_value, utime):
...
     if time_values[ceil_mask].size > 0:
         useconds = Unit('second')
-        second_frac = useconds.convert(0.75, utime.units)
-        dates[ceil_mask] = utime.num2date(time_values[ceil_mask] + second_frac)
+        second_frac = useconds.convert(0.75, time_units)
+        dates[ceil_mask] = cftime.num2date(
+            time_values[ceil_mask] + second_frac, **num2date_kwargs)
     dates[round_mask] = _discard_microsecond(dates[round_mask])
-    result = dates[0] if shape is () else dates.reshape(shape)
+    result = dates[0] if shape == () else dates.reshape(shape)
```
rcomer, line 531: These were giving me SyntaxWarnings (python3.9) so I thought I might as well change them while I'm in there.

cf_units/__init__.py
```diff
@@ -484,17 +394,17 @@ def _discard_microsecond(date):
...
     dates = dates.ravel()
-    # Create date objects of the same type returned by utime.num2date()
+    # Create date objects of the same type returned by cftime.num2date()
     # (either datetime.datetime or cftime.datetime), discarding the
     # microseconds
     dates = np.array([d and d.__class__(d.year, d.month, d.day,
                                         d.hour, d.minute, d.second)
                       for d in dates])
-    result = dates[0] if shape is () else dates.reshape(shape)
+    result = dates[0] if shape == () else dates.reshape(shape)
```
rcomer, line 403: These were giving me SyntaxWarnings (python3.9) so I thought I might as well change them while I'm in there.

rcomer:
Thanks @trexfeathers for your thorough review.  I think I've addressed everything.

cf_units/__init__.py
```diff
@@ -62,8 +62,6 @@
            'UT_ASCII',
            'FLOAT32',
            'FLOAT64',
-           'julian_day2date',
-           'date2julian_day',
```
trexfeathers, line 66: There are also entries in [utilities.rst](https://github.com/SciTools/cf-units/blob/master/doc/source/utilities.rst) that need removing.

rcomer: Done

trexfeathers (approved):

rcomer:
Thanks @trexfeathers! 🎉

Issue #171, opened 2021-06-08
Reboot cf-units


trexfeathers:
To-do checklist:
- [x] Enable GH discussions
  - [x] Create Releases GH discussions category
  - [x] Create `3.0` release discussion 
- [x] Add Issue templates
- [x] Add PR template
- [x] Create `3.0` release milestone
- [x] Create `3.0` project board
- [x] Update labels
- [x] Configure repo settings
  - [x] Allow squash merging only
  - [x] Require conversation resolution before merging
  - [x] Require linear history
- [x] Migrate CI from `travis-ci` to `cirrus-ci`
- [x] Adopt `setuptools-scm` and drop `versioneer`
- [x] Adopt `pre-commit`
  - [x] `black`
  - [x] `flake8`
  - [x] `isort`
- [x] Migrate `setup.py` to `setup.cfg`
- [x] Rationalise package dependencies for `conda` and PyPI
- [x] Update `README.md` badges
- [x] Remove `six` dependency
- [x] Simpler licence headers (e.g. SciTools/iris#3489)

PR #175, opened 2021-06-10
Switched from Travis to Cirrus. Updated requirements architecture.

trexfeathers:
### Description

Part of #171.

* Transfers CI from Travis to Cirrus.
* Switches coverage testing from Coveralls to Codecov.
* Also necessitates updating the dependency architecture to match the latest SciTools convention.
* Also removes the `six` dependency, which was missed during #157.

Plenty more #171 PR's to follow, but it's best to keep each PR 'atomic', and dropping Travis is the most time sensitive element.


requirements/py39.yml
```diff
@@ -0,0 +1,28 @@
+name: cf-units-dev
+
+channels:
+  - conda-forge
+
+dependencies:
+  - python 3.9
```
ocefpaf, line 7: Is there a way to create a build matrix with cirrus-ci? It seems a little brittle and hard to maintain  an env for each Python version.

trexfeathers: This is how ~50% of SciTools is set up now. @bjlittle is there more D.R.Y. way?

ocefpaf: Yeah, I'm seeing this in other projects. I wonder if going another direction with cirrus-ci is healthy for the projects in SciTools. It will be hard to get contributions from the larger SciPy community who is invested in GitHub Actions.

trexfeathers: A valid point. Although I would hope that most contributions shouldn't need involvement with the CI architecture.

The caching and compute credits are appealing parts of Cirrus - faster runs.

ocefpaf: > A valid point. Although I would hope that most contributions shouldn't need involvement with the CI architecture.

That is pretty much all I do since 2012!

> The caching and compute credits are appealing parts of Cirrus - faster runs.

GitHub Actions is quite fast, not sure if there are comparisons. Also, there are mechanism for caching too. I don't want to be pushy or anything. I just wonder if it is  going in another direction is worth it.

trexfeathers: I find this topic really interesting. I've also noticed that GHA is by far the most popular choice. And given that barriers to involvement are a big factor in the health of a project: we need a **really good** reason to go against the crowd.

From what you're saying, GHA at least has a chance of meeting the needs of SciTools projects, but we haven't tried it out yet, so at the moment we don't know if we have a really good reason to stick with Cirrus.

That said, the consensus is that SciTools and SciTools-incubator projects should have a consistent setup, so this conversation is bigger than this PR. I've therefore raised an issue against Iris - as the project with the largest test suite - to pursue this further: https://github.com/SciTools/iris/issues/4191. In the meantime this question shouldn't prevent this PR being merged IMHO.

rcomer: Looks like Cartopy is using GHA, so not completely alien to SciTools.

.cirrus.yml
```diff
@@ -0,0 +1,79 @@
...
+#   - https://cirrus-ci.org/guide/writing-tasks/
+#   - https://cirrus-ci.org/guide/linux/
+#   - https://hub.docker.com/_/gcc/
+#   - https://hub.docker.com/_/python/
+
+#
+# Global defaults.
+#
+container:
+  image: python:latest
```
bjlittle: @trexfeathers We could default to `gcc:latest` instead and then drop lines 43-44 in `linux_task`

.cirrus.yml
```diff
@@ -0,0 +1,79 @@
...
+      COVERAGE: "codecov"
+  name: "${CIRRUS_OS}: py${PY_VER}"
+  container:
+    image: gcc:latest
+  env:
+    PATH: ${HOME}/mambaforge/bin:${PATH}
+  mamba_cache:
+    folder: ${HOME}/mambaforge
+    fingerprint_script:
+      - wget --quiet https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-x86_64.sh -O mambaforge.sh
```
bjlittle, line 48: Just realised that this should be part of the `populate_script`, otherwise it's always pulling down mamba, and that's not always necessary

requirements/py37.yml
```diff
@@ -0,0 +1,28 @@
...
+channels:
+  - conda-forge
+
+dependencies:
+  - python 3.7
+
+# setup dependencies
+  - cython
+  - numpy
+  - pytest-runner
```
bjlittle: I don't *think* that we need this dependency, right?

setup.py
```diff
@@ -141,9 +141,20 @@ def _set_builtin(name, value):
...
+    install_requires=['antlr4-python3-runtime==4.7.2',
+                      'cftime>=1.2',
+                      'numpy',
+                      # udunits2 cannot be installed with pip, and it is
+                      #  expected to be installed separately.
+                      ],
     setup_requires=['pytest-runner', 'numpy', 'cython'],
-    tests_require=load('requirements-dev.txt'),
+    tests_require=['codecov',
+                   'cython',
```
bjlittle, line 152: The tests shouldn't explicitly need `cython`, as `pip install -e .` with the associated `pyproject.toml` will build the cython extensions and install `cf-units` in developer mode, and the tests will just run thanks to PEP517/518 :+1: 

I've tested this pattern with [SciTools-incubator/python-stratify](https://github.com/SciTools-incubator/python-stratify), which like `cf-units`, also used `cython` to build extensions - so you could see the pattern we've adopted there for fuller clarity.

bjlittle (changes requested):
@trexfeathers Looks good, just a few review comments to service.

Also, totally noted regarding everything concerning `cirrus-ci` vs GH actions. We can sort that out separately :+1:

requirements/py37.yml
trexfeathers: I'm willing to take your word on this. [Not the first time it has been questioned](https://github.com/conda-forge/cf_units-feedstock/blob/master/recipe/meta.yaml#L23), but I am yet to get involved with `py.test` so I was uncomfortable deciding this myself.

setup.py
trexfeathers: I'm well aware there's a lot more to do on architecture, but I've been trying to keep this PR as the minimum necessary to migrate the CI. Do you consider this one a deal breaker that can't wait?

.cirrus.yml
trexfeathers: e71ae3607e57ec75888fa55a9abfacce8bc9f880

.cirrus.yml
trexfeathers: Nice catch! 78a85ad79790133088dbbb3bcb06631dd4cb8c32

requirements/py37.yml
trexfeathers: 67ef259caf5f009c4b394020089d71d900ff07c4

.cirrus.yml
bjlittle: Ahhhh actually we do need it there, as the very next line does a `sha256sum` of it! Sorry, I'm being dense :stuck_out_tongue_closed_eyes:

setup.py
bjlittle: Cool. Totally understood. It can certainly wait, but it would be good to revisit this before we push another release :+1:

.cirrus.yml
```diff
@@ -0,0 +1,77 @@
...
+  env:
+    PATH: ${HOME}/mambaforge/bin:${PATH}
+  mamba_cache:
+    folder: ${HOME}/mambaforge
+    fingerprint_script:
+      - echo "${CIRRUS_OS} $(sha256sum mambaforge.sh)"
+      - echo "${MAMBA_CACHE_PACKAGES}"
+      - echo "$(date +%Y).$(expr $(date +%U) / ${CACHE_PERIOD}):${MAMBA_CACHE_BUILD}"
+    populate_script:
+      - wget --quiet https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-x86_64.sh -O mambaforge.sh
```
bjlittle: Sorry, my fault totally... we need to revert this change and include the `wget` as part of the `fingerprint_script` as line 48 is performing a `sha256sum mambaforge.sh` of the downloaded file, sorry.

.cirrus.yml
trexfeathers: I thought this ought to work - if the fingerprint errors doesn't it default to populating? Then next time it will be here?

[Just tested and it seems to work.](https://cirrus-ci.com/task/5524303732015104?logs=mamba#L1-8)

.cirrus.yml
trexfeathers: https://github.com/SciTools/cf-units/pull/175#discussion_r651028077

trexfeathers: 51a82523f0efe1ea2af1f9588395c61224a3980f

.cirrus.yml
trexfeathers: Ah I see now - this exists outside of the cache recovery, so not downloading the file during fingerprinting means it will _never_ be there. Have now reverted.

51a82523f0efe1ea2af1f9588395c61224a3980f

bjlittle (approved):
Looks good to go :+1:

trexfeathers:
Thanks @bjlittle!

PR #176, opened 2021-06-15
Setup architecture overhaul

trexfeathers:
### Description

More of #171 

* Migrate as much as possible from `setup.py` to `setup.cfg`.
* Use `pyproject.toml` where possible.
* Replace Versioneer with `setuptools_scm`.


trexfeathers:
I believe the coverage reduction is an artifact of marginally shrinking the codebase. Interested to hear the reviewer's thoughts on what we do:

* Merge with an apparent failure?
* Reconfigure Codecov to be marginally less sensitive?
* Reconfigure Codecov to not raise an error when coverage is reduced?

setup.cfg
```diff
@@ -1,32 +1,75 @@
...
+    cftime >=1.2
+    numpy
+    # udunits2 cannot be installed with pip, and it is expected to be
+    #  installed separately.
+packages = find:
+python_requires =
+    >=3.7
+setup_requires =
+    cython
+    numpy
```
bjlittle: @trexfeathers We now no longer need this `setup_requires` section, thanks to PEP517/518 support in the `pyproject.toml`

.coveragerc
```diff
@@ -7,9 +7,8 @@ branch = True
 plugins = Cython.Coverage
 include =
     cf_units/*
-omit = 
+omit =
     setup.py
-    versioneer.py
     cf_units/_version.py
     cf_units/etc/*
     cf_units/tests/*
```
bjlittle: @trexfeathers If we're in the business of minimising the number of configuration files, then `.coveragerc` can also be added to the `setup.cfg` :+1:

bjlittle: See https://coverage.readthedocs.io/en/coverage-5.0/config.html#configuration-reference

setup.py
```diff
@@ -15,7 +14,6 @@
 DEFINE_MACROS = None
 FLAG_COVERAGE = '--cython-coverage'  # custom flag enabling Cython line tracing
 BASEDIR = os.path.abspath(os.path.dirname(__file__))
```
bjlittle: @trexfeathers We could opt to use `pathlib` here instead of `os`, but I'm not greatly fussed :neutral_face:

bjlittle (changes requested):
@trexfeathers This looks really good.

Just a couple of minor suggestions to mull over, otherwise LGTM :+1:

bjlittle:
> I believe the coverage reduction is an artifact of marginally shrinking the codebase. Interested to hear the reviewer's thoughts on what we do:
> 
>     * Merge with an apparent failure?
> 
>     * Reconfigure Codecov to be marginally less sensitive?
> 
>     * Reconfigure Codecov to not raise an error when coverage is reduced?

@trexfeathers I'll look into this for you... we can configure this to behave how we want :+1:

bjlittle:
@trexfeathers Adding the following `codecov.yml` should do the trick:
```
coverage:
  status:
    project:
      default:
        target: auto     # auto compares coverage to the previous base commit
        threshold: 0.5%  # coverage can drop by up to 0.5% while still posting success
```
See https://docs.codecov.com/docs/common-recipe-list

bjlittle:
@trexfeathers I'd also nuke the `.stickler.yml` fine now in this PR - whilst we're here :wink: 

Also, I turned this service off on `stickler-ci.com` for SciTools/cf-units :+1:

trexfeathers:
>@trexfeathers Adding the following `codecov.yml` should do the trick:

@bjlittle what do you think of doing this in a [Team YAML](https://docs.codecov.com/docs/codecov-yaml#team-yaml)? Seems like a helpful pattern to adopt across all our uses of Codecov?

bjlittle:
> @bjlittle what do you think of doing this in a [Team YAML](https://docs.codecov.com/docs/codecov-yaml#team-yaml)? Seems like a helpful pattern to adopt across all our uses of Codecov?

yeah, for sure

trexfeathers:
I don't seem to have permissions to edit a Team YAML.

setup.cfg
trexfeathers: https://github.com/SciTools/cf-units/pull/176/commits/81aefb6c793fb403603d523dd8a28d00bbd1984a

.coveragerc
trexfeathers: a264e37069cb721ad1928cfe83419282214ade43
67e22b07599b0f0ca828fcea19a78bace1699c38
8547da42a68efd7e5e3a4b975ce8b589922c5d44

setup.py
trexfeathers: dd3d78045f80b46fd41d05eed494bedb0651ecc7

trexfeathers:
Back to you @bjlittle. Thanks for your patience

bjlittle:
> I don't seem to have permissions to edit a Team YAML.

@trexfeathers I'll take a look into this for you :+1:

setup.py
bjlittle: Above and beyond :rocket:

bjlittle (approved):

trexfeathers:
Thanks @bjlittle!

PR #177, opened 2021-06-18
Simple licence

trexfeathers:
### Description

Another item from #171.

In line with the latest [SciTools licensing guidelines](https://github.com/SciTools/scitools.org.uk/blob/master/documents/scitools_licensing.md#source-preamble):

* The licence headers are much shorter and mostly reference the main licensing files in the root directory.
  The years test in `test_coding_standards.py` has been replaced with a basic check that each file starts with the correct string.
* `README.md` aligns with the example in the guidelines.
* A `CONTRIBUTING.md` has been added, as this is referenced by the updated `README.md`.

PR #178, opened 2021-06-21
license headers follow up

rcomer:
### Description

There was a small omission in the license header test change at #177, so it couldn't produce a test failure.  I blame the reviewer.

Test failed for all files because of the initial new line in the template.  The `# -*- coding: utf-8 -*-` also caused a failure, but that was only needed at python 2 anyway.


bjlittle (approved):
LGTM :+1:

trexfeathers:
Thanks @rcomer!

PR #181, opened 2021-06-24
Implement Read the docs.

### LINKED ISSUE #169: Docs not up to date
The latest version of `cf_units` is currently 2.1.5.  The published docs appear to be only showing v2.0:
https://scitools.org.uk/cf-units/docs/latest/

trexfeathers:
### Description

Adds the necessary config to allow Read the docs to work. I will now set up Read the docs to build from the `main` branch.


requirements/py37.yml
```diff
@@ -25,3 +25,7 @@ dependencies:
 # dev dependencies
   - pip
   - pre-commit
+
+# docs dependencies
+  - setuptools_scm
```
bjlittle, line 30: This is more of a setup dependency even though it's used during docs building... just seems odd seeing it categorised here.

bjlittle (approved):

bjlittle:
Let's see how it goes... we can always tweak any changes as necessary.

But LGTM :+1:

requirements/py37.yml
trexfeathers: I agree. However without it the build was unable to pick up `cf_units.__version__`.

But when not using Read the Docs, everything works fine with it only listed in `pyproject.toml`!

trexfeathers:
Thanks @bjlittle! It seems to have built fine.

PR #183, opened 2021-06-29
Persist calendar when discarding cftime microsecond

bjlittle:
### Description
This PR addresses issues discovered whilst testing [SciTools/iris](https://github.com/SciTools/iris) when attempting to unpin `cftime`, whereby the specific `calendar` of `cftime.datetime` objects are not persisted whenever the `cf_units._discard_microsecond` function was called.

Once this PR is merged, I'd like to cut a `cf-units` 3.0.1 patch release to support an `iris` 3.0.3 patch release.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -406,17 +406,33 @@ def _discard_microsecond(date):
     dates = np.asarray(date)
     shape = dates.shape
     dates = dates.ravel()
+    result = None
     # Create date objects of the same type returned by cftime.num2date()
     # (either datetime.datetime or cftime.datetime), discarding the
     # microseconds
-    dates = np.array(
-        [
-            d
-            and d.__class__(d.year, d.month, d.day, d.hour, d.minute, d.second)
-            for d in dates
-        ]
-    )
-    result = dates[0] if shape == () else dates.reshape(shape)
+    discard = []
+    for dt in dates:
+        if dt:
+            kwargs = dict(
+                year=dt.year,
+                month=dt.month,
+                day=dt.day,
+                hour=dt.hour,
+                minute=dt.minute,
+                second=dt.second,
+            )
+            if isinstance(dt, cftime.datetime):
+                kwargs["calendar"] = dt.calendar
+            discard.append(dt.__class__(**kwargs))
+    if discard:
+        dates = np.array(discard)
+        if shape == ():
+            result = dates[0]
+        else:
+            if np.prod(shape) == dates.size:
+                result = dates.reshape(shape)
+            else:
+                result = dates
     return result
 
 
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -403,23 +403,14 @@ def _discard_microsecond(date):
         datetime, or numpy.ndarray of datetime object.
 
     """
-    dates = np.asarray(date)
+    dates = np.asanyarray(date)
     shape = dates.shape
     dates = dates.ravel()
-    result = None
-    # Create date objects of the same type returned by cftime.num2date()
-    # (either datetime.datetime or cftime.datetime), discarding the
-    # microseconds
-    discard = [dt.replace(microsecond=0) for dt in dates if dt]
-    if discard:
-        dates = np.array(discard)
-        if shape == ():
-            result = dates[0]
-        else:
-            if np.prod(shape) == dates.size:
-                result = dates.reshape(shape)
-            else:
-                result = dates
+
+    # using the "and" pattern to support masked arrays of datetimes
+    dates = np.array([dt and dt.replace(microsecond=0) for dt in dates])
+    result = dates[0] if shape == () else dates.reshape(shape)
+
     return result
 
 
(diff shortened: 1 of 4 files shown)


cf_units/__init__.py
```diff
@@ -406,17 +406,33 @@ def _discard_microsecond(date):
...
-        [
-            d
-            and d.__class__(d.year, d.month, d.day, d.hour, d.minute, d.second)
-            for d in dates
-        ]
-    )
-    result = dates[0] if shape == () else dates.reshape(shape)
+    discard = []
+    for dt in dates:
+        if dt:
```
bjlittle: I've honoured this short-circuit logic here, although I must confess I don't completely appreciate why we have to do this... I'm sure it's historical.

cf_units/tests/unit/unit/test_Unit.py
```diff
@@ -67,7 +67,7 @@ def test_non_gregorian_calendar_conversion_dtype(self):
             (np.float64, True),
             (np.int32, False),
             (np.int64, False),
-            (np.int, False),
+            (int, False),
```
bjlittle, line 70: See https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

pyproject.toml
```diff
@@ -62,6 +62,7 @@ exclude = '''
     | build
     | dist
   )/
+  | _version.py
```
bjlittle, line 65: This file is auto-generated by `setuptools-scm` and should not be linted

cf_units/__init__.py
```diff
@@ -406,17 +406,33 @@ def _discard_microsecond(date):
...
+                year=dt.year,
+                month=dt.month,
+                day=dt.day,
+                hour=dt.hour,
+                minute=dt.minute,
+                second=dt.second,
+            )
+            if isinstance(dt, cftime.datetime):
+                kwargs["calendar"] = dt.calendar
+            discard.append(dt.__class__(**kwargs))
```
rcomer: Wondering if we could get there more simply with the `replace` method.
https://docs.python.org/3.8/library/datetime.html#datetime.datetime.replace
https://unidata.github.io/cftime/api.html?highlight=datetime#cftime.datetime.replace

```python
discard.append(dt.replace(microsecond=0))
```

rcomer: `cftime.datetime` also has a `has_zero_year` attribute, though I don't know if that matters for Iris.

bjlittle: Nice. I'll give that a try

cf_units/__init__.py
```diff
@@ -406,17 +406,20 @@ def _discard_microsecond(date):
...
     # microseconds
-    dates = np.array(
-        [
-            d
-            and d.__class__(d.year, d.month, d.day, d.hour, d.minute, d.second)
-            for d in dates
-        ]
-    )
-    result = dates[0] if shape == () else dates.reshape(shape)
+    discard = [dt.replace(microsecond=0) for dt in dates if dt]
```
rcomer: OK I done a bit of digging.  I **think** the `d and` bit is so we can handle masked arrays, and went in at #68.  If we have a masked array of datetimes

```python
unit = cf_units.Unit('days since 1970-01-01')
dates = unit.num2date(np.ma.masked_array(range(5), mask=[0, 0, 1, 1, 0]))
```

If we just do

```python
[d.replace(microsecond=0) for d in dates]
```
we get an `AttributeError`.  If instead we do

```python
[d and d.replace(microsecond=0) for d in dates]
```
we get our masked array back.
```
[cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 0, 0, has_year_zero=False),
 cftime.DatetimeGregorian(1970, 1, 2, 0, 0, 0, 0, has_year_zero=False),
 masked,
 masked,
 cftime.DatetimeGregorian(1970, 1, 5, 0, 0, 0, 0, has_year_zero=False)]
```

Which is different from using `if d`, as that reduces the size of the array

```python
[d.replace(microsecond=0) for d in dates if d]
```
```
[cftime.DatetimeGregorian(1970, 1, 1, 0, 0, 0, 0, has_year_zero=False),
 cftime.DatetimeGregorian(1970, 1, 2, 0, 0, 0, 0, has_year_zero=False),
 cftime.DatetimeGregorian(1970, 1, 5, 0, 0, 0, 0, has_year_zero=False)]
```

So I **think** we should still be using `dt and`.

rcomer (commented):
Hi @bjlittle I have one comment for your consideration.  I've not yet looked at the test module as it seems likely that the outcome of the comment may change the tests.

cf_units/__init__.py
bjlittle: Cool. Thanks for the context... I'll change and extend the tests to cover this.

pyproject.toml
rcomer: Full disclosure: I don't understand this file.  But if it's auto-generated and @bjlittle is happy with it then it should be 👍

rcomer (approved):

PR #186, opened 2021-07-07
Build and publish Wheels

ocefpaf:
Nothing to see here for now. Just testing...

### DIFF
--- a/.github/workflows/actions/entrypoint.sh
+++ b/.github/workflows/actions/entrypoint.sh
@@ -1,19 +1,20 @@
 #!/bin/sh
 
 yum install -y udunits2-devel
-PYTHONS=("cp36-cp36m" "cp37-cp37m" "cp38-cp38" "cp39-cp39")
+PYTHONS=("cp37-cp37m" "cp38-cp38" "cp39-cp39")
+WHEELHOUSE="/github/workspace/wheelhouse/"
 
 cp /usr/include/udunits2/* /usr/include/
 
-SITECFG=cf_units/etc/site.cfg
-echo "[System]" > $SITECFG
-echo "udunits2_xml_path = /usr/share/udunits/udunits2.xml" >> $SITECFG
+export UDUNITS2_INCDIR="/usr/include/udunits2"
+export UDUNITS2_LIBDIR="/usr/lib64"
+export UDUNITS2_XML_PATH="/usr/share/udunits/udunits2.xml"
 
 for PYTHON in ${PYTHONS[@]}; do
     /opt/python/${PYTHON}/bin/pip install --upgrade pip wheel setuptools setuptools_scm build twine auditwheel
-    /opt/python/${PYTHON}/bin/python -m build --sdist --wheel . --outdir /github/workspace/wheelhouse/
+    /opt/python/${PYTHON}/bin/python -m build --sdist --wheel . --outdir ${WHEELHOUSE}
 done
 
-for whl in /github/workspace/wheelhouse/cf_units*.whl; do
+for whl in ${WHEELHOUSE} cf_units*.whl; do
     auditwheel repair $whl
 done
--- a/.github/workflows/actions/entrypoint.sh
+++ b/.github/workflows/actions/entrypoint.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp36-cp36m" "cp37-cp37m" "cp38-cp38" "cp39-cp39")
+
+cp /usr/include/udunits2/* /usr/include/
+
+SITECFG=cf_units/etc/site.cfg
+echo "[System]" > $SITECFG
+echo "udunits2_xml_path = /usr/share/udunits/udunits2.xml" >> $SITECFG
+
+for PYTHON in ${PYTHONS[@]}; do
+    /opt/python/${PYTHON}/bin/pip install --upgrade pip wheel setuptools setuptools_scm build twine auditwheel
+    /opt/python/${PYTHON}/bin/python -m build --sdist --wheel . --outdir /github/workspace/wheelhouse/
+done
+
+for whl in /github/workspace/wheelhouse/cf_units*.whl; do
+    auditwheel repair $whl
+done
--- a/.github/workflows/actions/entrypoint.sh
+++ b/.github/workflows/actions/entrypoint.sh
@@ -4,8 +4,6 @@ yum install -y udunits2-devel
 PYTHONS=("cp37-cp37m" "cp38-cp38" "cp39-cp39")
 WHEELHOUSE="/github/workspace/wheelhouse/"
 
-cp /usr/include/udunits2/* /usr/include/
-
 export UDUNITS2_INCDIR="/usr/include/udunits2"
 export UDUNITS2_LIBDIR="/usr/lib64"
 export UDUNITS2_XML_PATH="/usr/share/udunits/udunits2.xml"
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -1,0 +1,40 @@
+name: linux-wheels
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  packages:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Get tags
+      shell: bash
+      run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+
+    - uses: ./.github/workflows/actions/manylinux2010_x86_64
+    - name: copy manylinux wheels
+      run: |
+        mkdir dist
+        cp wheelhouse/cf_units*-manylinux2010_x86_64.whl dist/
+
+    - name: CheckFiles
+      shell: bash
+      run: |
+        ls dist
+    
+    - name: Test wheels
+      shell: bash
+      run: |
+        python -m pip install wheel twine pytest
+        cd dist && python -m pip install cf_units*.whl && python -m pytest --pyargs cf_units
+        python -m twine check *
+
+    - name: Publish a Python distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_PASSWORD }}
--- a/.github/workflows/actions/manylinux2010_x86_64/action.yml
+++ b/.github/workflows/actions/manylinux2010_x86_64/action.yml
@@ -1,0 +1,7 @@
+name: 'build wheels with manylinux2010_x86_64'
+description: 'build wheels with manylinux2010_x86_64'
+runs:
+    using: 'docker'
+    image: docker://quay.io/pypa/manylinux2010_x86_64
+    args:
+        - .github/workflows/actions/entrypoint.sh


bjlittle:
@ocefpaf I was about to crank out the `bdist_wheel`s for `cf-units` manually... but if we can automate this, it would be lovely :star_struck:

.github/workflows/actions/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp36-cp36m" "cp37-cp37m" "cp38-cp38")
+
+for PYTHON in ${PYTHONS[@]}; do
+    /opt/python/${PYTHON}/bin/pip install --upgrade pip wheel setuptools setuptools_scm pep517 twine auditwheel
+    /opt/python/${PYTHON}/bin/python -m pep517.build --source --binary . --out-dir /github/workspace/wheelhouse/
```
bjlittle: @ocefpaf Is https://github.com/pypa/build mature and stable enough yet to use instead?

ocefpaf: It is. I changed that in my next commit. This was copy-n-pasta from another repo.

ocefpaf:
@bjlittle I'm not going to work on Windows and macOS at the moment b/c those require some extra knowledge and time that I don't have. (Installing udunits2 in those systems, finding the right paths, etc.)

This is almost ready but there is a problem with the `udunits2.xml` file. We rely on site.cfg and a fixed path for it. For the wheel we should copy that file with the distribution and change the code to find it. Or I'm just not thinking of a nice pythonic way to define an env var for site.cfg that aims into the installed directory.

TL;DR I'm not going to dig into this b/c code changes here would imply extra tests and adaptations to avoid breakages.

Do you have any insights on how to solve this?

.github/workflows/actions/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp36-cp36m" "cp37-cp37m" "cp38-cp38" "cp39-cp39")
+
+cp /usr/include/udunits2/* /usr/include/
+
+SITECFG=cf_units/etc/site.cfg
+echo "[System]" > $SITECFG
+echo "udunits2_xml_path = /usr/share/udunits/udunits2.xml" >> $SITECFG
```
ocefpaf: This doesn't work. See https://github.com/SciTools/cf-units/pull/186#issuecomment-876801650 for a longer explanation.

ocefpaf: Note to sell: pyproj does this in a sane manner. We can probably use a similar approach.

ocefpaf: We are getting somewhere in https://github.com/SciTools/cf-units/pull/186 and the wheel should make this easier.

bjlittle: @ocefpaf Okay, I'll take a peek and see what I can do :+1:

bjlittle: @ocefpaf Hmmm I guess we could push the creation of the `site.cfg` down into the `setup.py`. At least then it would make it all more consistent and self-contained.

Although this seems to be delaying the inevitable of needing to know the `share` install directory of `udunits2`, which ultimately needs to be provided i.e., this doesn't seem to be easily discoverable and correct with any confidence outside `conda` packaging, whereas the install location of `udunits2` is known within `conda` thanks to the `udunits2` activation/deactivation scripts setting/unsetting `UDUNITS2_XML_PATH`.

My only suggestion is that we:

- push the creation of the `site.cfg` down into the `setup.py`
- the `setup.py` enforces that the `UDUNITS2_XML_PATH` is set, otherwise `cf-units` won't build/install

This means that:
- the `cf-units` recipe `bld.bat` and `bld.sh` are simplified for `conda` (thanks to the `udunits2` package setting `UDUNITS2_XML_PATH`
- for `pip` the `UDUNITS2_XML_PATH` requires to be explicitly set beforehand

What do you think?

I'll push a PR to do this if you think it's a wise strategy to adopt.... :thinking:

ocefpaf: So users would be responsible for setting UDUNITS2_XML_PATH in pip from source and wheels, right? I'm OK with the former, so users can have the flexibility of building from any udunits2 they want,  but not the latter. For the latter to "just work" we would need something like pyproj, where we copy the xml file into the `.lib` directory. Does that make sense?

bjlittle: @ocefpaf You're referring to https://github.com/pyproj4/pyproj/blob/master/setup.py#L191:L206, using https://pyproj4.github.io/pyproj/stable/installation.html#pyproj-build-environment-variables, right?

ocefpaf: Yep.

ocefpaf: @bjlittle, crazy idea, what if we rely on `UDUNITS2_XML_PATH` for the conda package as you said above and do a "download if unset+cache" from a static GitHub raw link? Like cartopy's data?

dopplershift: Not sure if it's helpful, but we are posting a single combined XML file made directly from the UDUNITS-2 repo [here](https://docs.unidata.ucar.edu/thredds/udunits2/current/udunits2_combined.xml).

bjlittle: @ocefpaf Sorry for dragging this out... I've been busy enjoying myself at SciPy2021 :wink: 

Okay, so I *think* I might have a ~hack~ way around this. I'm testing in a manylinux docker container, but if this hangs together then I'll push a separate PR, and then you can simply rebase and we're cookin' with gas.

Essentially, we can follow the `pyproj` trick... it seems like a reasonable enough approach to me. Similarly, we can have a `CFUNITS_WHEEL` environment variable which acts as a hook to `setup.py` to instruct it to inject additional `package_data` that consists of the `/usr/share/udunits/*.xml` files (or file, thanks @dopplershift). This will be copied by `setup.py` to `cf_units/etc/share/udunits/*.xml` and shipped in the `wheel`. But this only all happens if the `wheel` generated  `.lib` directory exists as well.

Also, the additional step needed here now is that if no `cf_units/etc/site.cfg` exists **AND** the `cf_units/etc/share/...` exists, then it was a `wheel` install, and `cf-units` can complete the configuration to the XML files at runtime, as it has them bundled and it knows where they are relative to itself (resolved absolutely, hence no need to know the location pre-install, as that's troublesome)... otherwise all bets are off.

I'd opt to bundle the `xml` file/s not in the `.lib` directory, as that seems a bit abusive of the `wheel` mechanism - unless you think/recommend otherwise?

As we move forwards, we could adopt this pattern for `conda` too, which means that we could ultimately drop the use of `build.sh` and `build.bat`, right?

The only downside to this is that for legal reasons it's not cool to bundle the `xml` files with `cf-units` as a `wheel`, but I'm hoping that either of you might have thoughts on that... and it's not a blocker.

bjlittle: > crazy idea, what if we rely on `UDUNITS2_XML_PATH` for the conda package as you said above and do a "download if unset+cache" from a static GitHub raw link? Like cartopy's data?

I love crazy... many wonderful things have happened as a result of brief madness.

The only downside to this I can think of is no internet (I'm thinking of a poor oceanographer stuck in a metal can floating on the high seas in the middle of no where attempting to install `cf-units`... now that is crazy)... but bundling the `xml` with `cf-units` is attractive, since it's now self contained all in the wheel :thinking:

ocefpaf: Been there, done that. Forgot to download cartopy's data before a cruise and it was not fun :smile:

bjlittle: @ocefpaf Okay, I think I've cracked it. I'm kinda happy with a reasonably neat pattern now.

I'll test it this evening, and if I'm happy push a separate PR, which I'll ping you on :+1:

bjlittle: The good news is that I've finally manged to create a manylinux bdist wheel which contains the `xml` files:
```
File Name                                             Modified             Size
cf_units-3.0.2.dev0+g5aa890a.d20210717.dist-info/WHEEL 2021-07-17 00:37:36          148
cf_units-3.0.2.dev0+g5aa890a.d20210717.dist-info/top_level.txt 2021-07-17 00:30:54            9
cf_units-3.0.2.dev0+g5aa890a.d20210717.dist-info/RECORD 2021-07-17 00:37:36         4670
cf_units-3.0.2.dev0+g5aa890a.d20210717.dist-info/COPYING 2021-07-17 00:31:02        35147
cf_units-3.0.2.dev0+g5aa890a.d20210717.dist-info/METADATA 2021-07-17 00:31:02         6023
cf_units/_udunits2.pxd                         2021-07-16 09:51:12         2571
cf_units/_udunits2.pyx                         2021-07-16 09:51:12         8554
cf_units/config.py                             2021-07-16 15:55:16         1843
cf_units/util.py                               2021-07-16 09:51:12         3265
cf_units/__init__.py                           2021-07-16 09:52:44        61342
cf_units/_udunits2.c                           2021-07-17 00:30:50       665160
cf_units/_version.py                           2021-07-17 00:30:54          196
cf_units/_udunits2.cpython-38-x86_64-linux-gnu.so 2021-07-17 00:37:36      1002928
cf_units/tex.py                                2021-07-16 09:51:12         1325
cf_units/etc/site.cfg.template                 2021-07-16 09:50:32           51
cf_units/etc/share/udunits2.xml                2021-07-17 00:30:54          519
cf_units/etc/share/udunits2-prefixes.xml       2021-07-17 00:30:54         2304
cf_units/etc/share/udunits2-common.xml         2021-07-17 00:30:54        85083
cf_units/etc/share/udunits2-base.xml           2021-07-17 00:30:54         2772
cf_units/etc/share/udunits2-derived.xml        2021-07-17 00:30:54         8594
cf_units/etc/share/udunits2-accepted.xml       2021-07-17 00:30:54        11617
cf_units/tests/test_unit.py                    2021-07-16 09:51:12        32490
cf_units/tests/__init__.py                     2021-07-16 09:51:12          201
cf_units/tests/test_coding_standards.py        2021-07-16 09:51:12         4586
cf_units/tests/test_tex.py                     2021-07-16 09:51:12          895
cf_units/tests/integration/test_num2date.py    2021-07-16 09:51:12          760
cf_units/tests/integration/__init__.py         2021-07-16 09:51:12          201
cf_units/tests/integration/test__num2date_to_nearest_second.py 2021-07-16 09:51:12        11619
cf_units/tests/integration/test_num2pydate.py  2021-07-16 09:51:12          889
cf_units/tests/integration/test_date2num.py    2021-07-16 09:51:12         2511
cf_units/tests/integration/parse/test_parse.py 2021-07-16 09:51:12         7640
cf_units/tests/integration/parse/test_graph.py 2021-07-16 09:51:12         1640
cf_units/tests/unit/__init__.py                2021-07-16 09:51:12          251
cf_units/tests/unit/test__udunits2.py          2021-07-16 09:51:12         8819
cf_units/tests/unit/test__discard_microsecond.py 2021-07-16 09:52:44         3941
cf_units/tests/unit/unit/test_as_unit.py       2021-07-16 09:51:12         2282
cf_units/tests/unit/unit/test_Unit.py          2021-07-16 09:52:44         9704
cf_units/_udunits2_parser/README.md            2021-07-16 09:51:12         1528
cf_units/_udunits2_parser/__init__.py          2021-07-16 09:51:12         7226
cf_units/_udunits2_parser/udunits2Parser.g4    2021-07-16 09:51:12         1987
cf_units/_udunits2_parser/.gitignore           2021-07-16 09:51:12           62
cf_units/_udunits2_parser/graph.py             2021-07-16 09:51:12         3275
cf_units/_udunits2_parser/compile.py           2021-07-16 09:51:12         2464
cf_units/_udunits2_parser/udunits2Lexer.g4.jinja 2021-07-16 09:51:12         3618
cf_units/_udunits2_parser/parser/udunits2Lexer.py 2021-07-16 09:51:12        26495
cf_units/_udunits2_parser/parser/udunits2Parser.py 2021-07-16 09:51:12        37254
cf_units/_udunits2_parser/parser/udunits2ParserVisitor.py 2021-07-16 09:51:12         2152
cf_units/_udunits2_parser/parser/__init__.py   2021-07-16 09:51:12            0
cf_units.libs/libudunits2-3783a498.so.0.1.0    2021-07-17 00:37:34       122880
```

bjlittle: But I have one outstanding question...

During the build process I was under the impression that a `cf_units/.lib` directory would be created, and that is an indicator within `setup.py` that a wheel is being built.

This isn't the case. Are  you able to explain? Why is this the case for `pyproj` and not for `cf-units`? Or am I misunderstanding or doing something wrong :thinking: 

Is the `.lib` directory specific to `pyproj` ?

ocefpaf: The `.lib` I believe is created only during the autidwheel phase. I may be mistaken though, I'm definitely not an expert.

bjlittle: @ocefpaf fancy taking on #191 ?

Bank that to support this PR 👍

bjlittle: @ocefpaf If you now rebase your PR with `main`, given that #191 is merged, I think you need to replace `lines 6-10` with the following:
```bash
export UDUNITS2_INCDIR="/usr/include/udunits2"
export UDUNITS2_LIBDIR="/usr/lib64"
export UDUNITS2_XML_PATH="/usr/share/udunits/udunits2.xml"
```

.github/workflows/actions/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp36-cp36m" "cp37-cp37m" "cp38-cp38" "cp39-cp39")
```
bjlittle: @ocefpaf We've now dropped support for `cp36` 🎉 

See https://github.com/SciTools/cf-units/blob/main/setup.cfg#L48

.github/workflows/actions/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+cp /usr/include/udunits2/* /usr/include/
+
+SITECFG=cf_units/etc/site.cfg
+echo "[System]" > $SITECFG
+echo "udunits2_xml_path = /usr/share/udunits/udunits2.xml" >> $SITECFG
+
+for PYTHON in ${PYTHONS[@]}; do
+    /opt/python/${PYTHON}/bin/pip install --upgrade pip wheel setuptools setuptools_scm build twine auditwheel
+    /opt/python/${PYTHON}/bin/python -m build --sdist --wheel . --outdir /github/workspace/wheelhouse/
```
bjlittle: You could factor `/github/workspace/wheelhouse/` into an environment variable, as it's also used in `line 17`.

Just a suggestion, not really fussed, your choice 👍

bjlittle:
@ocefpaf Test away my man... ping me when you think this PR is good to go.

Take it out of draft and I'll re-review again... it's mighty close now 🍻

ocefpaf:
Screenshot from 2021-07-28 20-58-07


PS: I'll work on macOS and Windows at a later time. Linux was already quite the battle for now :smile:

.github/workflows/actions/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp37-cp37m" "cp38-cp38" "cp39-cp39")
+WHEELHOUSE="/github/workspace/wheelhouse/"
+
+cp /usr/include/udunits2/* /usr/include/
```
bjlittle: @ocefpaf Do we still need this `cp`?

I was hoping that the new `UDUNITS2_INCDIR` and `UDUNITS2_LIBDIR` addressed this 🤔

ocefpaf: Probably yes. I forgot to test it. Doing it now and if it works I'll remove it.

ocefpaf: It works and the latest commit remove that part. Thanks!

.github/workflows/actions/entrypoint.sh
ocefpaf: Done!

.github/workflows/actions/entrypoint.sh
ocefpaf: Removed.

.github/workflows/build_linux.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        python -m pip install wheel twine pytest
+        cd dist && python -m pip install cf_units*.whl && python -m pytest --pyargs cf_units
+        python -m twine check *
+
+    - name: Publish a Python distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_PASSWORD }}
```
bjlittle, line 40: I created an API token on PyPI for a generic `scitools-pypi` user that we have, and also created the associated `cf-units` GitHub repository secret `PYPI_PASSWORD`

.github/workflows/build_linux.yml
```diff
@@ -0,0 +1,40 @@
...
+    
+    - name: Test wheels
+      shell: bash
+      run: |
+        python -m pip install wheel twine pytest
+        cd dist && python -m pip install cf_units*.whl && python -m pytest --pyargs cf_units
+        python -m twine check *
+
+    - name: Publish a Python distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@master
```
bjlittle, line 37: We should pin this to a specific stable release version, say `@release/v1.4.2` which is the latest.

But I'll do that as a separate PR and ping you on that... I think that you've been through enough already

pypi-publich-github-action

bjlittle (approved):
Awesome, let's go for it @ocefpaf :rocket: 

I'll create a follow-up PR to pin the action release version used :+1:

PR #194, opened 2021-07-30
3p0px update

bjlittle:
### Description
Update the `v3.0.x` release feature branch with updates from `main`.

Note that, the merge-back from `v3.0.x` to `main` (#192) was squash merged rather than done with a merge commit, therefore the `v3.0.x` commit `5aa890a` has been re-hashed as `d9f280c` - I'm not going to bother unpicking this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -27,7 +27,7 @@ env:
 #
 # Linting
 #
-lint_task:
+task:
   auto_cancellation: true
   container:
     image: python:latest
@@ -44,7 +44,7 @@ lint_task:
 #
 # Testing (Linux)
 #
-linux_task:
+task:
   auto_cancellation: true
   matrix:
     env:
--- a/.github/workflows/actions/entrypoint.sh
+++ b/.github/workflows/actions/entrypoint.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+yum install -y udunits2-devel
+PYTHONS=("cp37-cp37m" "cp38-cp38" "cp39-cp39")
+WHEELHOUSE="/github/workspace/wheelhouse/"
+
+export UDUNITS2_INCDIR="/usr/include/udunits2"
+export UDUNITS2_LIBDIR="/usr/lib64"
+export UDUNITS2_XML_PATH="/usr/share/udunits/udunits2.xml"
+
+for PYTHON in ${PYTHONS[@]}; do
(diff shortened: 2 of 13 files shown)


bjlittle:
@ocefpaf Fancy taking this on?

Please ensure to do a **merge commit**, rather than a **squash commit** 👍

This'll update the `v3.0.x` release feature branch, at which point I'll cut a post patch release and we can see whether the GH PyPI publish action is behaving :+1:

ocefpaf:
Sure.

> Please ensure to do a **merge commit**, rather than a **squash commit** 

Oh. I thought you set that up and liked it that way, I really dislike it.
 
> This'll update the `v3.0.x` release feature branch, at which point I'll cut a post patch release and we can see whether the GH PyPI publish action is behaving

:+1:

ocefpaf (approved):

bjlittle:
@ocefpaf okay, so I'm testing the GH action after seeing initial failures.

I'm at the point where the wheels successfully build. However, installing the wheels for the appropriate python versions and running the tests results in some failures.... Which I'm investigating 🤔

ocefpaf:
I can take a look later next week. We needed one GHA merged to get it to work, now it is easier to debug.

bjlittle:
@ocefpaf I've finally got it working!

There were a few hoops to jump through, and I think that I've finally settled on a reasonable GHA workflow, see https://github.com/bjlittle/cf-units/actions/runs/1086268122

I also had to make a couple of changes to `cf-units`, but nothing major. However, the biggest gotcha that stumped me was working out that the `manylinux2010` is too old a platform to build the `cf-units` wheels on as the `yum install -y udunits2-devel` installed `udunits2 2.1.19-1`... which is ~10 years old and causes some serious parsing errors in `cf-units` that the test suite was catching - and I couldn't work around them :cry: 

However, I've re-built the wheels based on `manylinux2014` (see [here](https://github.com/pypa/manylinux#manylinux) for the differences), which in turn installs `udunits2 2.2.20-2` circa 2015, and that's modern enough to successfully build `py37`, `py38` and `py39` with all the tests passing :partying_face: 

I'll push a PR to SciTools/cf-units today or tomorrow. BTW it also includes the `sdist`.

If you're interested my branch is https://github.com/bjlittle/cf-units/tree/test-action and I was testing with this release https://github.com/bjlittle/cf-units/releases/tag/v3.0.1.post0

I may now extend to cover `macos`, but I'll see how I'm doing for time

PR #195, opened 2021-07-30
fix gh action version

bjlittle:
### Description
The `pypa/gh-action-pypi-publish` suggests the [syntax](https://github.com/pypa/gh-action-pypi-publish#usage) for the GH action version, but clearly this isn't the case.

Trying again...


rcomer:
New syntax is consistent with actions elsewhere 👍

Looks like `release/v1` is actually a [branch](https://github.com/pypa/gh-action-pypi-publish/tree/release/v1).

PR #196, opened 2021-08-02
Fix GHA pypi-publish and UDUNITS2 XML read

bjlittle:
### Description
This PR builds on the fine work started by @ocefpaf (#186) to automate the building and publishing of `cf-units` wheels to PyPI using GHA.

We require to repair the binary wheels at a minimum as `manylinux2014` wheels, rather than `manylinux2010` wheels. This is due to the bad version of `udunits2` `2.1.19-1` installed by `yum install -y udunits2-devel` within the older `manylinux2010` docker container, see [here](https://github.com/SciTools/cf-units/pull/194#issuecomment-890431077) for further details.

See the successful run of this GHA on my [cf-units fork](https://github.com/bjlittle/cf-units/actions/runs/1090088865).

Note that, this PR only supports building and publishing `linux` wheels and the `sdist`.

I'm keen to bank this PR to unblock https://github.com/ioos/compliance-checker/pull/867. Afterwards, we can look at attempting to adopt the [cibuildwheel](https://github.com/pypa/cibuildwheel) GHA workflow, if possible. It appears to do a lot of the heavy lifting that this PR re-invents, and it also provides automated wheel testing and other platform wheel building coverage, which is very attractive :+1: 

TODO:
- [x] Push a separate commit to this PR that folds in the PyPI upload step.


cf_units/tests/unit/test__udunits2.py
```diff
@@ -26,7 +27,10 @@ class Test_get_system(unittest.TestCase):
     """
 
     def test_read_xml(self):
-        system = _ud.read_xml()
+        try:
+            system = _ud.read_xml()
+        except _ud.UdunitsError:
+            system = _ud.read_xml(get_xml_path())
```
bjlittle, line 33: This pattern requires to be adopted within `test__udunits2.py` in order to successfully test a `wheel` installation.

setup.cfg
```diff
@@ -40,6 +40,7 @@ include_package_data = True
 install_requires =
     antlr4-python3-runtime ==4.7.2
     cftime >=1.2
+    jinja2
```
bjlittle, line 43: This is a top level dependency, so let's be explicit rather than relying on `antlr4` pulling it in.

Required when auto-testing the wheels install.

ocefpaf (approved):

cf_units/__init__.py
```diff
@@ -187,15 +185,8 @@ def suppress_errors():
...
-        _alt_xml_path = config.get_option(
-            "System",
-            "udunits2_xml_path",
-            default=os.path.join(
-                sys.prefix, "share", "udunits", "udunits2.xml"
-            ),
-        )
         try:
-            _ud_system = _ud.read_xml(_alt_xml_path.encode())
+            _ud_system = _ud.read_xml(config.get_xml_path())
```
bjlittle, line 189: `codecov` `patch` was complaining about this line change, which is pretty impossible to explicitly test since it's a statement at the `cf_units` root module level... so I turned `patch` `off`, which is a reasonable stance to take.

The behaviour here of `cf-units` is pretty fundamental to the `wheels` installation, and this line is being exercised successfully on `import` after a wheels install, otherwise no tests would run (nor would `cf-units` import) based on a wheels package install. Also, we know that GHA testing of a wheels installation is good, and normal CI testing is good, so I'm okay with this.

bjlittle:
@ocefpaf I think this is now good to go... although, I was slightly cheeky and turned `codecov` `patch` `off`.

This is simple because `codecov` `patch` was complaining about  a `cf_units` module level statement that I'd rationalised and not explicitly tested... sadly it's pretty impossible to test what it was complaining about.

That said, I'm satisfied that `cf_units` is behaving as we expect without explicitly testing the change. We have normal CI testing and GHA wheels testing to testify to this :+1:

bjlittle:
@ocefpaf I'd love to see this merged... I'm super curious to see whether the workflow hangs together :smile:

ocefpaf:
Sorry. I'm teaching the OceanHackWeek this week, so I'll be a bit slow.

Looks like the only option available to merge is the one you do not want, right?
Or should we use squash and merge?

Screenshot from 2021-08-03 21-47-09

bjlittle:
@ocefpaf Good luck and enjoy OceanHackWeek :rocket: 

Yeah, our team typically `squash and merge` for normal pull requests to keep a clean linear history, and use a `merge commit` only when merging back a feature branch to `main`. Meh.

So it's all good. Go for it mate, if you're happy to merge :+1:

bjlittle:
@ocefpaf Awesome! It worked! 

See:
- https://github.com/SciTools/cf-units/actions/runs/1097577715
- https://pypi.org/project/cf-units/3.0.1.post0/#files

Thanks for all your help and your patience :heart:

Issue #185, opened 2021-07-02
For modern data allow unit conversion between standard and proleptic_gregorian calendars


larsbarring:
## ✨ Feature Request
For modern data allow unit conversion between `standard` and `proleptic_gregorian` calendars. "Modern data" is data sources that do not span the Julian/Gregorian transition.

## Motivation
When using iris to analyse data from different sources, or to build a multi-model ensemble cube, it often happens that one input source is based on the `standard` calendar, and another on the `proleptic_gregorian` calendar. For data that spans the Julian/Gregorian transition these calendars are clearly not compatible, and conversion is of course not possible. 

From the perspective of the producer of the data there might be very good reasons for choosing one or the other calendar. E.g. a modelling centre might want to have only one calendar irrespective of whether a simulation covers only the Gregorian era or spans the Julian/Gregorian transition. And a centre dealing with observational data would want to rely on a calendar that follows the real world calendar even for the very longest time-series. But from the analyst's perspective, when fusing modern datasets it should be easy to harmonise the calendars. 

After all the `standard` calendar is the *mixed Julian/Gregorian calendar*, and the `proleptic_gregorian` calendar is the *Gregorian calendar extended backwards in time*. That is, for the modern [post Julian] period they are the same calendar.

EDIT: In principle one could argue that a corresponding conversion should be possible between the `standard` calendar and the `Julian` calendar, but I imagine there is much less demand for this because the vast majority of relevant datasets cover the modern period.

larsbarring:
@rcomer, following up on the comments in PR #215 I found that all this might be **much simplified** by just using [`cftime.datetime.change_calendar`](https://unidata.github.io/cftime/api.html). On the other hand, it seems that cf-units is not applying any sanity checks to the date of the unit string (#216). And this may cause problem here.

Is there already a string decoder somewhere in scitools capable of handling "CF-formatted" date/time strings, i.e. not zero-padded and not making assumptions about which millennium?

rcomer:
Nice, I didn't know about the `change_calendar` method, and I agree that looks like a good approach.  I'm not sure about a string decoder, but maybe it would be easiest to get the reference date as a `datetime` object by

```python
ref_date = u.num2date(0)
```
I tried this with your examples from #216, and the `num2date` step does seem to throw error for most of the cases it should.  Does that help?

larsbarring:
I am closing this issue because already some time ago the requested functionality was implemented.  Thanks @rcomer !

PR #205, opened 2021-11-05
Support gregorian to standard calendar change

### LINKED ISSUE #203: Change definitive calendar form from "gregorian" to "standard"
## 📰 Custom Issue
As of CF version 1.9 the `gregorian` calendar is deprecated. This means that the *definitive form* of this calendar should be `standard` (not the other way around).

larsbarring:
### Description
Change definitive form of calendar from `gregorian` to `standard`, instead of the other way around. 
With first time attempt at [change of ] unit tests for this.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -130,7 +130,7 @@ CALENDARS = [
 #: Where calendars have multiple names, we map the alias to the
 #: definitive form.
 CALENDAR_ALIASES = {
-    CALENDAR_STANDARD: CALENDAR_GREGORIAN,
+    CALENDAR_GREGORIAN: CALENDAR_STANDARD,
     CALENDAR_NO_LEAP: CALENDAR_365_DAY,
     CALENDAR_ALL_LEAP: CALENDAR_366_DAY,
 }
@@ -830,7 +830,7 @@ class Unit(_OrderedHashable):
                 raise value_error from None
             if _OP_SINCE in unit.lower():
                 if calendar is None:
-                    calendar_ = CALENDAR_GREGORIAN
+                    calendar_ = CALENDAR_STANDARD
                 elif isinstance(calendar, (str,)):
                     calendar_ = calendar.lower()
                     if calendar_ in CALENDAR_ALIASES:
@@ -1815,7 +1815,7 @@ class Unit(_OrderedHashable):
             # gregorian calendar as it handles these and udunits does not.
             if (
                 self.is_time_reference()
-                and self.calendar != CALENDAR_GREGORIAN
+                and self.calendar != CALENDAR_STANDARD
             ):
                 result_datetimes = cftime.num2date(
                     result, self.cftime_unit, self.calendar
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1813,10 +1813,7 @@ class Unit(_OrderedHashable):
                 result = copy.deepcopy(value)
             # Use cftime for converting reference times that are not using a
             # gregorian calendar as it handles these and udunits does not.
-            if (
-                self.is_time_reference()
-                and self.calendar != CALENDAR_STANDARD
-            ):
+            if self.is_time_reference() and self.calendar != CALENDAR_STANDARD:
                 result_datetimes = cftime.num2date(
                     result, self.cftime_unit, self.calendar
                 )
--- a/cf_units/tests/test_unit.py
+++ b/cf_units/tests/test_unit.py
@@ -47,7 +47,7 @@ class Test_unit__creation(unittest.TestCase):
 
     def test_no_calendar(self):
         u = Unit("hours since 1970-01-01 00:00:00")
-        self.assertEqual(u.calendar, unit.CALENDAR_GREGORIAN)
+        self.assertEqual(u.calendar, unit.CALENDAR_STANDARD)
 
     def test_unsupported_calendar(self):
         with self.assertRaisesRegex(ValueError, "unsupported calendar"):
@@ -576,9 +576,9 @@ class Test_stringify(unittest.TestCase):
 
     def test___repr___time_unit(self):
         u = Unit(
-            "hours since 2007-01-15 12:06:00", calendar=unit.CALENDAR_STANDARD
+            "hours since 2007-01-15 12:06:00", calendar=unit.CALENDAR_GREGORIAN
         )
-        exp = "Unit('hours since 2007-01-15 12:06:00', calendar='gregorian')"
+        exp = "Unit('hours since 2007-01-15 12:06:00', calendar='standard')"
         self.assertEqual(repr(u), exp)
 
 
@@ -949,11 +949,11 @@ class TestTimeEncoding(unittest.TestCase):
 class TestNumsAndDates(unittest.TestCase):
     def test_num2date(self):
         u = Unit(
-            "hours since 2010-11-02 12:00:00", calendar=unit.CALENDAR_STANDARD
+            "hours since 2010-11-02 12:00:00", calendar=unit.CALENDAR_GREGORIAN
         )
         res = u.num2date(1)
         self.assertEqual(str(res), "2010-11-02 13:00:00")
-        self.assertEqual(res.calendar, "gregorian")
+        self.assertEqual(res.calendar, "standard")
         self.assertIsInstance(res, cftime.datetime)
 
     def test_num2date_py_datetime_type(self):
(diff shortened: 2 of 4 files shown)


trexfeathers:
Thanks for submitting this @larsbarring! I was taking a look and I wondered if the changes to `__init__` really would be that simple - I'm glad you think the same 🙂

larsbarring:
Yeah, but it remains to be seen. The unit tests fails miserably (I am a complete rookie at these). And I have no insight regarding possible downstream dependencies

trexfeathers:
>The unit tests fails miserably (I am a complete rookie at these). And I have no insight regarding possible downstream dependencies

The traceback is the same error we have been seeing recently during documentation builds, so we can definitely say it's not resulting in changes you have made. Unfortunately I am also clueless to the fix at the moment!

trexfeathers:
I've asked if @bjlittle could take a look at this - the problem was first observed at 5540e08ec88d0669bcc4071d931a7f96e9e026a9. But we're quite limited on time at the moment so there might be a wait.

larsbarring:
Thanks, I have something for #185 in the pipeline, but it uses this change.

larsbarring:
@trexfeathers @bjlittle Any chance the [unit test problem](https://github.com/SciTools/cf-units/pull/205#issuecomment-964287627) might be fixed reasonably soon? That is, soon enough that the upcoming iris 3.2.0 will use cf-units with this fix implemented.

trexfeathers:
> @trexfeathers @bjlittle Any chance the [unit test problem](https://github.com/SciTools/cf-units/pull/205#issuecomment-964287627) might be fixed reasonably soon? That is, soon enough that the upcoming iris 3.2.0 will use cf-units with this fix implemented.

We've seen your message 🙂. We'll have a better idea of the answer in the next week.

wjbenfold:
Rebasing onto main should resolve the unit test problem @larsbarring

trexfeathers:
>Rebasing onto main should resolve the unit test problem @larsbarring

I've just been checking locally and there are some _actual_ unit test failures relating to the changes, but they look fairly simple 🙂 We'll see the detail once this is rebased.

larsbarring:
The linting failure I am not sure what to do about. And now two of the three initial test failures have been fixed. The third one is a bit more complicated. It seems that in `Units.__num2date_to_nearest_second` cftime returns an instance of `cftime.DatetimeGregorian` despite that the calendar specified in the call is `"standard"`, cf. cftime [issue #256](https://github.com/Unidata/cftime/issues/256) and [PR #257](https://github.com/Unidata/cftime/pull/257). Maybe this could be an issue elsewhere where cftime is called?

rcomer:
> The linting failure I am not sure what to do about

I think maybe we need to do for `cf_units` what https://github.com/SciTools/iris/pull/4235 did for Iris.  That's a bit outside my area of competency though!

trexfeathers:
> The linting failure I am not sure what to do about

@larsbarring are you using pre-commit/black locally? I know we haven't made this expectation obvious.

larsbarring:
hrm ... probably not as it seems, at least not correctly configured ....

trexfeathers:
@larsbarring if you install pre-commit in your development environment I _think_ it should "just work" whenever you make a commit 🤞

larsbarring:
Hm, this is becoming a bit too complex. The pre-commit black gives the same uninformative error, and seems to have reformatted lines that I have not touched L. 846-849, as well as L 1816-1819 where I changed the calendar.

trexfeathers:
I see what @rcomer is saying now. I've asked if anyone in the team has some time to look at this.

Thanks, as ever, for your patience @larsbarring 😬

trexfeathers:
> > The linting failure I am not sure what to do about
> 
> I think maybe we need to do for `cf_units` what [SciTools/iris#4235](https://github.com/SciTools/iris/pull/4235) did for Iris. That's a bit outside my area of competency though!

Done, thanks to @wjbenfold ❤. @larsbarring if you bring in the latest `main` this ought to fix the linting trouble.

larsbarring:
Linting failure solved -- thanks @wjbenfold, @trexfeathers, @rcomer !

The remaining test failure is because of Unit.num2date returns gregorian calendar:
> in Units.__num2date_to_nearest_second cftime returns an instance of cftime.DatetimeGregorian despite that the calendar specified in the call is "standard",

Here is a quick hack to show the problem:

```python
import cftime
import cf_units as unit
Unit = unit.Unit
u = Unit(
   "hours since 2010-11-02 12:00:00", calendar=unit.CALENDAR_GREGORIAN
)
print("repr(u)      =  ", repr(u))
res = u.num2date(1)
print("repr(res)    =  ", repr(res))
print("res.calendar =  ", res.calendar)

# repr(u)      =   Unit('hours since 2010-11-02 12:00:00', calendar='standard')
# repr(res)    =   cftime.DatetimeGregorian(2010, 11, 2, 13, 0, 0, 0, has_year_zero=False)
# res.calendar =   gregorian
```
And I do not know whether to wait for `cftime` to do something, or how fix this within `iris` in a safe and general way.

rcomer:
As far as I can tell, it's impossible to create a standard calendar datetime instance with the current `cftime` release:

``` 
In [4]: cftime.datetime(2022, 1, 1, calendar="standard")
Out[4]: cftime.datetime(2022, 1, 1, 0, 0, 0, 0, calendar='gregorian', has_year_zero=False)

In [5]: test = _

In [6]: test.calendar = "standard"
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-fc68e10f067c> in <module>
----> 1 test.calendar = "standard"

AttributeError: attribute 'calendar' of 'cftime._cftime.datetime' objects is not writable
```
So I think the only way to get `num2date` to return a standard calendar object is to wait for the next `cftime` release, which will include the changes from https://github.com/Unidata/cftime/pull/257.  At that point I think it will change to "standard" regardless of what we do here.

Perhaps it doesn't matter.  Since you are looking at the CF Conventions, I guess your main concern is how NetCDF files are written?  I have an existing NetCDF file that was saved by a previous Iris version.  The time coordinate looks like this with `ncdump`:

```
	double time(forecast_reference_time) ;
		time:bounds = "time_bnds" ;
		time:units = "hours since 1970-01-01 00:00:00" ;
		time:standard_name = "time" ;
		time:calendar = "gregorian" ;
```

I made an environment with your branch and Iris v3.1.  I used this to load and re-save my file.  Now the `ncdump` shows me this:

```
	double time(forecast_reference_time) ;
		time:bounds = "time_bnds" ;
		time:units = "hours since 1970-01-01 00:00:00" ;
		time:standard_name = "time" ;
		time:calendar = "standard" ;
```

So for NetCDF saving, the change seems to be having the desired effect.  So one option may be to revert the expected value of the `num2date` test and bank the rest of the change.

There may be other considerations that I am missing.  What do others think?

larsbarring:
My main reason for pushing this issue, except for the general argument that it is nice to follow  CF, is that we need to concatenate data from different sources that happens to have different calendars (#185). But in a PR to solve that issue I could include tests of both calendars (that later could compacted once a new version of cftime is available).

rcomer:
@SciTools/cf-units-devs I believe these tests should now pass with the latest version of `cftime`.  I tried re-running the py39 tests this morning but that didn't help.  Is there a way to make the tests start again with a fresh environment?

bjlittle:
TBH not sure we can get this banked in time for `iris` 3.2 (which is our main focus at the moment), but we can swarm on it afterwards, as we need to make sure we get this right.

If banking this PR forces subsequent changes upstream within `iris`, then we can opt to release an `iris` 3.2.1 patch to align everything.

Sound like a reasonable plan?

wjbenfold:
> I believe these tests should now pass with the latest version of `cftime`. I tried re-running the py39 tests this morning but that didn't help. Is there a way to make the tests start again with a fresh environment?

It looks like if a py3.9 lockfile already exists, it refuses to overwrite it even if the build number gets bumped, so I'm not sure @rcomer

wjbenfold:
Ok, I think if you rebase onto main then we can hopefully get the CI passing @larsbarring. Then we can move to reviewing with confidence that what's here passes the tests

larsbarring:
Pretty nice to see this passing the checks. Thanks all, for helping me with setting up these these tools :-)

bjlittle (approved):
@larsbarring Awesome! Thanks so much :100: :beers:

wjbenfold:
> If banking this PR forces subsequent changes upstream within `iris`, then we can opt to release an `iris` 3.2.1 patch to align everything.

Note for the interested - having run the Iris tests with current cf-units main (with this change) and current Iris, the tests all pass. We can jump up and down on it a bit more after the next cf-units release to double-check but it's looking good :)

rcomer:
I've also tried running the Iris tests with this and _did_ get failures, so not sure what we're doing differently.  The good news is those failures were easy to fix with find-and-replace within Iris.

PR #207, opened 2021-11-12
Fix py310 test

ocefpaf:
I'm not sure how to set cirrus-ci and if I added the new env file correctly. I'm not familiar with this multiple files setup. Hopefully things are in the right place.

### DIFF
--- a/requirements/py38.yml
+++ b/requirements/py38.yml
@@ -12,7 +12,7 @@ dependencies:
 
 # core dependencies
   - antlr-python-runtime 4.7.2.*
-  - cftime>=1.2,<1.5.1.1
+  - cftime>=1.2,<=1.5.1.1
   - numpy
   - udunits2
 
--- a/requirements/py38.yml
+++ b/requirements/py38.yml
@@ -12,7 +12,7 @@ dependencies:
 
 # core dependencies
   - antlr-python-runtime 4.7.2.*
-  - cftime>=1.2
+  - cftime>=1.2,<1.5.1.1
   - numpy
   - udunits2
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,6 +52,8 @@ task:
     env:
       PY_VER: "3.9"
       COVERAGE: "codecov"
+    env:
+      PY_VER: "3.10"
   name: "${CIRRUS_OS}: py${PY_VER}"
   env:
     PATH: ${HOME}/mambaforge/bin:${PATH}
--- a/.github/workflows/actions/entrypoint.sh
+++ b/.github/workflows/actions/entrypoint.sh
@@ -6,7 +6,7 @@ set -e
 
 yum install -y udunits2-devel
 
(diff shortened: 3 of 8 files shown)


ocefpaf:
BTW, I'm not able to build latest release or main with modern pip/setuptools. I'm not sure what changed but I can only build the release before 3.0.1.post0. See https://github.com/conda-forge/cf_units-feedstock/pull/45

trexfeathers:
>BTW, I'm not able to build latest release or main with modern pip/setuptools. I'm not sure what changed but I can only build the release before 3.0.1.post0.

Our Read the Docs builds show that this first occurred at commit 5540e08ec88d0669bcc4071d931a7f96e9e026a9.

requirements/py38.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
 
 # core dependencies
   - antlr-python-runtime 4.7.2.*
-  - cftime>=1.2
+  - cftime>=1.2,<1.5.1.1
```
ocefpaf: We probably don't want this pin but I'm not sure what is the right fix for cftime new default. Should I just change the expected calendar in that test? If so, the test is a bit flaky and ensuring the expected calendar explicitly is probably better than relaying on cftime's defaults.

ocefpaf:
Python 3.10 is failing now b/c we are missing the older cftime on that py version. I added it to conda-forge and this test should pass when we restart. However, we should probably just update the test to the latest cftime and pin to cftime>1.5.1.1 instead. What do you all think?

requirements/py38.yml
rcomer: Over at #205, @larsbarring actively wants to change the expected calendar in that test.  But now I’m confused, as I thought https://github.com/Unidata/cftime/pull/257 would be needed to change it, and `cftime`‘s GitHub page is not indicating any releases since September 😕

larsbarring: I was in contact with cftime developers in [Unidata/cdtime#256](https://github.com/Unidata/cftime/issues/256) and initially they did a required change in [Unidata/cdtime#257](https://github.com/Unidata/cftime/pull/257). But it it broke xarray, and thus was not merged into main. So I [asked the  xarray devs](https://github.com/pydata/xarray/discussions/6162) and they have some ideas. But I am already way out of my competence zone, so I will just have to wait see --- and hope that I have not caused any trouble for you  guys.

rcomer: > hope that I have not caused any trouble for you guys.

Not at all, I think by opening #205 you have helped us prepare for this change whereas usually we catch up some time later.

I am just confused because there seems to be a new `cftime` release that is not advertised on their GitHub page (which is where I had been checking).

ocefpaf: @rcomer maybe we can take the conservative approach for now and issue one more release with a cftime pin to unlock a  py310 wheel?

bjlittle:
Replaced by #214

PR #209, opened 2022-01-14
Improve precommit config to apply fixes locally

wjbenfold:
### Description

See #205 for context. Implements something similar to https://github.com/SciTools/iris/pull/4235 for cf-units

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -36,9 +36,7 @@ task:
   name: "${CIRRUS_OS}: precommit hooks"
   precommit_script:
     - python -m pip install --retries 3 pre-commit
-    - pre-commit run --all-files black
-    - pre-commit run --all-files flake8
-    - pre-commit run --all-files isort
+    - SKIP="no-commit-to-branch" pre-commit run --all-files
 
 
 #
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,8 +33,8 @@ task:
     image: python:latest
     cpu: 2
     memory: 4G
-  name: "${CIRRUS_OS}: linting"
-  lint_script:
+  name: "${CIRRUS_OS}: precommit hooks"
+  precommit_script:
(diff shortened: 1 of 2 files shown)


trexfeathers:
~~@wjbenfold have you been able to work out that cf-units doesn't have any paths to exclude like Iris does? That was the main thing I was worried about.~~

https://github.com/SciTools/cf-units/pull/209#issuecomment-1013047744

wjbenfold:
I'm not confident of that, no. So far just happy that it appears to be doing the same things at the same time as I'd expect from Iris.

Are there any other things you'd like checking that I might not have thought of (by which I mean anything at all that you'd check)?

trexfeathers:
Scratch what I said before! It wasn't excluded files it was **excluded hooks** that I was thinking of.

>Are there any other things you'd like checking that I might not have thought of (by which I mean anything at all that you'd check)?

1. Renaming stuff that says `linting` to `pre-commit`, to make it clear that it's doing whatever pre-commit is doing.
2. Make sure the list of pre-commit hooks that get run excludes those that would be harmful/non-sensical.

Sorry I see now that I jumped the gun and hadn't realised you were still working through the PR 😳

wjbenfold:
Definitely my bad for not making it draft. I'll get on those things now

wjbenfold:
> Make sure the list of pre-commit hooks that get run excludes those that would be harmful/non-sensical.

The only one that Iris skips and that thus seems necessary to skip is `no-commit-to-branch`. I've told the CI to do that, and in the CI pre-commit check it shows up as skipped, but the pre-commit.ci check shows is Passed rather than Skipped. Is the pre-commit stuff running twice?

I've just checked, and the same things happens in Iris' CI, so I don't think it's a bad thing here

trexfeathers (approved):
Great stuff @wjbenfold , thanks so much!

And I can confirm that this is doing what I would expect, too.

PR #213, opened 2022-02-09
Make mamba overwrite the existing lockfile on cache miss

wjbenfold:
### Description

Adding --yes option to `mamba create` command in the cirrus ci config so that it can overwrite an existing lockfile if the cache missed


rcomer:
The calendar failure is expected due to the new cftime version.  #207 proposes a pin; #205 updates the expected value.

wjbenfold:
@bjlittle what's the best route forward here? I can't have the tests pass without #205 and #205 can't have the tests pass without this... 🐔🥚

bjlittle (approved):
@wjbenfold Cyclic shenanigans... what joy.

Okay, let's just bank this and it's a step forward in the right direction :+1:

PR #214, opened 2022-02-16
infrastructure refresh

bjlittle:
### Description
This PR shows some love to the `cf-units` infra-structure.

In a nutshell:
- replaces `cirrus-ci` with GHA
- adopts use of GH `dependabot`
- adopts the use of `tox`
- adopts the automated building of `conda-lock` explicit environment spec files
  - supports `linux`, `osx` and `windows` locks  
- retires the previous hand crafted PyPI `sdist` and `bdist` building/testing/publishing and replaced with the `cibuildwheel` GHA
- refreshes `readthedocs` to use `mamba` for faster building
- adopts the `setuptools-scm-git-archive` plugin
- uses the `oldest-supported-numpy` meta package
- fixes the tests to support `py310` (replaces/incorporates #207)
- introduces min pin of `py38`
- moves the documentation from the `doc` directory to the `docs` directory

Post merge, suggested follow-on tasks to consider would be to:
- extend `ci-wheels` to build for `osx` and `windows` using `cibuildwheel`, and
- rationalise the GHAs to use `composite` actions.

This ones for you @ocefpaf :beers: :wink:


trexfeathers:
If no-one has looked at this by Monday I will carve out some time.

ocefpaf (dismissed):

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+      matrix:
+        os: [ubuntu-latest]
+        version: [py38, py39, py310]
+        include:
+          - os: ubuntu-latest
+            platform: linux
+          - os: ubuntu-latest
+            version: py310
+            posargs: "--cov-report=xml --cov"
+            post-command: codecov
```
trexfeathers, line 39: The `codecov` command doesn't seem to be able to reach CodeCov - [see this run](https://github.com/SciTools/cf-units/runs/5284809124?check_suite_focus=true#step:10:519).

tox.ini
```diff
@@ -0,0 +1,56 @@
...
+setenv =
+    LOCKDIR = {toxinidir}{/}requirements{/}locks
+    TMPFILE = {envtmpdir}{/}cf-units.yml
+    YMLFILE = {toxinidir}{/}requirements{/}cf-units.yml
+skip_install =
+    true
+commands =
+    cp {env:YMLFILE} {env:TMPFILE}
+    python -c 'from sys import version_info as v; fh = open("{env:TMPFILE}", "a"); fh.write(f"\n  - python =\{v.major\}.\{v.minor\}\n")'
+    conda-lock --channel conda-forge --file {env:TMPFILE} --platform linux-64 --filename-template "{envname}-\{platform\}.txt" {posargs}
```
trexfeathers: This currently falls over. We are experiencing similar issues everywhere we run conda-lock. I can't provide my traceback as it includes proprietary information, but I expect the latest on our investigation can be found here: SciTools/iris#4638.

pyproject.toml
```diff
@@ -43,10 +44,11 @@ testpaths = "cf_units"
 
 [tool.setuptools_scm]
 write_to = "cf_units/_version.py"
+local_scheme = "dirty-tag"
 
 [tool.black]
 line-length = 79
-target-version = ['py37', 'py38', 'py39']
+target-version = ["py38", "py39"]
```
trexfeathers, line 52:
```suggestion
target-version = ["py38", "py39", "py310"]
```

Unless there's a reason this was omitted here?

setup.cfg
```diff
@@ -46,19 +43,18 @@ install_requires =
...
 docs =
-    setuptools_scm
     sphinx
 test =
     codecov
     cython
     jinja2
     pip
-    pytest
+    pytest>=6.0
```
trexfeathers: This pin isn't reflected elsewhere. What's it for? Should it be included against other mentions of `pytest`?

tox.ini
```diff
@@ -0,0 +1,56 @@
...
+    linux|darwin
+setenv =
+    LOCKDIR = {toxinidir}{/}requirements{/}locks
+    TMPFILE = {envtmpdir}{/}cf-units.yml
+    YMLFILE = {toxinidir}{/}requirements{/}cf-units.yml
+skip_install =
+    true
+commands =
+    cp {env:YMLFILE} {env:TMPFILE}
+    python -c 'from sys import version_info as v; fh = open("{env:TMPFILE}", "a"); fh.write(f"\n  - python =\{v.major\}.\{v.minor\}\n")'
```
trexfeathers, line 31: Much neater than maintaining multiple identical YAML files 😎

.pre-commit-config.yaml
```diff
@@ -1,8 +1,16 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+
+files: |
+    (?x)(
+        setup\.py|
+        cf_units\/.+\.py
+    )
```
trexfeathers: Couldn't you just replace this with a global `types: [file, python]`?

[This seems to be encouraged by the pre-commit docs](https://[pre-commit](https://pre-commit.com/#filtering-files-with-types).com/#filtering-files-with-types):

>Filtering with `types` provides several advantages over traditional filtering with `files`.
>
> * no error-prone regular expressions
> * ...

And as far as I can tell this would achieve the same thing as the pattern you've written here, unless I've misunderstood the pattern?

And I suppose there might also be a deliberate reason you haven't used the `types` key for the Black item?

.github/workflows/actions/entrypoint.sh
```diff
@@ -1,37 +0,0 @@
-#!/usr/bin/env bash
```
trexfeathers, line 1: I have no experience with cf-units' advanced build process, so I'm relying on @ocefpaf's approval here 🙂

.github/workflows/actions/manylinux2014_x86_64/action.yml
```diff
@@ -1,7 +0,0 @@
-name: 'build wheels with manylinux2014_x86_64'
```
trexfeathers, line 1: I have no experience with cf-units' advanced build process, so I'm relying on @ocefpaf's approval here 🙂

.github/workflows/ci-wheels.yml
```diff
@@ -0,0 +1,126 @@
...
+          python -m pytest --pyargs cf_units
+        UDUNITS2_INCDIR: /usr/include/udunits2
+        UDUNITS2_LIBDIR: /usr/lib64
+        UDUNITS2_XML_PATH: /usr/share/udunits/udunits2.xml
+        CIBW_ENVIRONMENT_PASS_LINUX: UDUNITS2_INCDIR UDUNITS2_LIBDIR UDUNITS2_XML_PATH
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: pypi-artifacts
+        path: ${{ github.workspace }}/wheelhouse/*.whl
```
trexfeathers, line 58: I have no experience with cf-units' advanced build process, so I'm relying on @ocefpaf's approval here 🙂

pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [build-system]
 # Defined by PEP 518
 requires = [
-  "setuptools>=42",
-  "setuptools_scm[toml]>=6.0",
+  "setuptools>=45",
+  "setuptools_scm[toml]>=6.2",
+  "setuptools_scm_git_archive",
```
trexfeathers, line 6: I understand that this extra plugin is needed to make sure that `setuptools_scm` gets the right version information.

.github/workflows/ci-wheels.yml
```diff
@@ -0,0 +1,126 @@
...
+#   - https://github.com/actions/download-artifact
+#   - https://github.com/actions/upload-artifact
+#   - https://github.com/pypa/cibuildwheel
+#   - https://github.com/pypa/build
+#   - https://github.com/pypa/gh-action-pypi-publish
+#   - https://test.pypi.org/help/#apitoken
+
+name: ci-wheels
+
+on: [pull_request, push]
```
trexfeathers:
```suggestion
on:
  pull_request:

  push:
    branches-ignore:
      - "conda-lock-auto-update"
      - "pre-commit-ci-update-config"
      - "dependabot/*"
```

See bjlittle/geovista#50.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,99 @@
+# Reference:
+#   - https://github.com/actions/cache
+#   - https://github.com/actions/checkout
+#   - https://github.com/conda-incubator/setup-miniconda
+
+name: ci-tests
+
+on: [pull_request, push]
```
trexfeathers:
```suggestion
on:
  pull_request:

  push:
    branches-ignore:
      - "conda-lock-auto-update"
      - "pre-commit-ci-update-config"
      - "dependabot/*"
```

See bjlittle/geovista#50.

.github/workflows/ci-locks.yml
```diff
@@ -0,0 +1,134 @@
...
+      # v3.12.1 (release date 31 Jan 2022)
+      uses: peter-evans/create-pull-request@f22a7da129c901513876a2380e2dae9f8e145330
+      with:
+        add-paths: ${{ github.workspace }}/requirements/locks/*.txt
+        commit-message: "updated conda lock files"
+        branch: conda-lock-auto-update
+        delete-branch: true
+        title: "[cf-units.ci] conda lock auto-update"
+        body: |
+          Conda lock files auto-updated to latest resolved environment for `cf-units` dependencies.
```
trexfeathers, line 123:
```suggestion
          Conda lock files auto-updated to latest resolved environment for `cf-units` dependencies.
          
          ### If the full suite of PR checks have not run, please close and re-open this pull request to trigger them.
```

See bjlittle/geovista#51.

trexfeathers (changes requested):
That was a tough review!

Thanks for the effort @bjlittle - a big improvement. Got some comments for you.

(@bjlittle and I suspect that having `setuptools` in the dependencies YAML means that the other setup dependencies (specified in `pyproject.toml`) will all get picked up automatically. If not, we will find out when RTD fails!)

pyproject.toml
bjlittle: As yet, it's not supported and including it causes an issue

tox.ini
bjlittle: It's a *bit* of a hack, but I adopted this pattern on `geovista` as I did find it irksome having multiple identical yaml files, apart from the `python` version for `iris`.

bjlittle: @trexfeathers If we done the same on `iris`, would this upset any assumptions that you've built into the `asv` benchmarking proces?

trexfeathers: > @trexfeathers If we done the same on `iris`, would this upset any assumptions that you've built into the `asv` benchmarking proces?

It'll probably be fine. Environment management is [this block](https://github.com/SciTools/iris/blob/b141b01b754fb0f012a0b4ffc812d66a3ca4c271/benchmarks/asv.conf.json#L21-L24) and [this block](https://github.com/SciTools/iris/blob/b141b01b754fb0f012a0b4ffc812d66a3ca4c271/noxfile.py#L361-L375). Looking at those, switching to Tox wouldn't exactly be a find-and-replace job, but we'd just need an easy way of instructing Tox to build an environment for **only the latest** Python.

pyproject.toml
trexfeathers: In that case it's rather odd that we're testing against Python `v3.10` - easy to misunderstand. Please could we either:

* Remove other references to `v3.10`?
  OR
* Stick an explanatory note here?

.github/workflows/ci-wheels.yml
bjlittle: @trexfeathers It's pretty much equivalent to this existing line in [cf-units/.github/workflows/actions/entrypoint.sh](https://github.com/SciTools/cf-units/blob/main/.github/workflows/actions/entrypoint.sh#L37)

pyproject.toml
bjlittle: @trexfeathers You've got to love open-source development... `py310` is available now :rocket:

setup.cfg
bjlittle: It was a min pin for the major release, but there's no special need... so I'll remove it :+1:

.pre-commit-config.yaml
bjlittle: TBH this was inherited from `iris`, and in that case we definitely wanted to use that pattern as we didn't want to check all `*.py` files.

As you point out, quite rightly, this doesn't apply for `cf-units` (as far as I can tell) so I'll remove it and see what behaviour we get.

Also, from what I can recall, the choice of pattern for `black` was due to the way that we were using it in CI... and again on reflection, that might not hold here... so let's remove it and see how it rolls

tox.ini
bjlittle: Added a fix, which is to opt for `--kind explicit`

.github/workflows/ci-tests.yml
bjlittle: hmmmm yeah... seen this on geovista, then once i merged it, it worked :neutral_face:

bjlittle: i've added the secret to explicitly point to codecov... so, try and see?

dunno

.github/workflows/ci-wheels.yml
bjlittle: @ocefpaf already added his approval to this pr 7 days ago...

.pre-commit-config.yaml
bjlittle: Seems fine to me 👍

.github/workflows/ci-tests.yml
trexfeathers: [Still no good](https://github.com/SciTools/cf-units/runs/5641443060?check_suite_focus=true#step:10:519).

>seen this on geovista, then once i merged it, it worked 😐

OK, I'll ignore this one then.

trexfeathers (approved):

trexfeathers:
Beautiful @bjlittle!

And the PR's not bad either 😁

PR #215, opened 2022-02-16
Issue #185: Convert calendars if possible wrt Julian-Gregorian gap

### LINKED ISSUE #185: For modern data allow unit conversion between standard and proleptic_gregorian calendars

larsbarring:
### Description
**Test and possible proof-of-concept**
Takes care of issue #185, and also allows conversion between `julian` and `standard` calendars if all datetimes are before 1582-10-5 0:0:0

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -135,6 +135,10 @@ CALENDAR_ALIASES = {
     CALENDAR_ALL_LEAP: CALENDAR_366_DAY,
 }
 
+#: Calendars that possibly could be converted into each other
+#: depending on min/max date
+GREGORIAN_CALENDARS = {CALENDAR_STANDARD, CALENDAR_PROLEPTIC_GREGORIAN}
+JULIAN_CALENDARS = {CALENDAR_STANDARD, CALENDAR_JULIAN}
 
 #
 # floating point types
@@ -962,7 +966,12 @@ class Unit(_OrderedHashable):
             True
 
         """
-        return self.calendar is not None
+        try:
+            result = self.calendar is not None
+        except ValueError:
+            result = False
+        print(result)
+        return result
 
     def is_long_time_interval(self):
         """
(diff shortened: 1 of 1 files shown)


larsbarring:
Ready for review -- with cftime 1.5.2 all test pass. But this I leave to you.

rcomer:
Thanks for this @larsbarring, it looks like it should be useful new functionality.  However I _think_ it may be necessary and sufficient to check the reference date in the unit, rather than checking the value dates themselves.  For example, if we have two units with references after the "gap" and inspect some date values before the gap, we have:

```python
import cf_units

modern_standard = cf_units.Unit("days since 1582-10-20", calendar="standard")
modern_proleptic_gregorian = cf_units.Unit("days since 1582-10-20", calendar="proleptic_gregorian")

print("standard:", modern_standard.num2date(-20))
print("proleptic_gregorian:", modern_proleptic_gregorian.num2date(-20))
```
```
standard: 1582-09-20 00:00:00
proleptic_gregorian: 1582-09-30 00:00:00
```
So the standard/julian date is 10 days before the proleptic gregorian one, but [these two dates represent the same actual day](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar#Difference_between_Julian_and_proleptic_Gregorian_calendar_dates), so the conversion seems valid.

Conversely if our reference dates are before the "gap" but our values are after:

```python
ancient_standard = cf_units.Unit("days since 1582-10-01", calendar="standard")
ancient_proleptic_gregorian = cf_units.Unit("days since 1582-10-01", calendar="proleptic_gregorian")

print("standard:", ancient_standard.num2date(20))
print("proleptic_gregorian:", ancient_proleptic_gregorian.num2date(20))
```
we actually get two different dates:

```
standard: 1582-10-31 00:00:00
proleptic_gregorian: 1582-10-21 00:00:00
```

What do you think?

larsbarring:
Thanks for this @rcomer ! -- I hadn't thought along these lines, which of course are much simpler. One possible snag would be that the Julian and proleptic Gregorian calendars have different methods for defining leap years (the 400 year difference). But when in each calendar this difference kicks in I am not sure.

larsbarring:
@rcomer Having now mulled over your idea a bit more I think that you are right! And that  it is even more general than what I was considering. 

In the real world continuum of time a *day* is just a convenient and useful unit. And the real world calendars (`standard`, `julian` and `proleptic_gregorian`) are just three alternative methods for labelling these days; sometimes the labels overlap and sometimes not. As the calendars, i.e. the "labelling methods", are well-known is should be possible to convert between them (at least for dates onwards from 1-1-1). 

Is this a fair interpretation of what you suggest?

rcomer:
@larsbarring, yes that's what I'm thinking - the calendars are different labels for the same actual sequence of days.  I have to admit I hadn't begun to consider what happens at or before 1-1-1 though!

larsbarring:
@rcomer, from your first comment I take it that you might already have a PR on its way? Anyway, I will close this PR because because following this approach will result in rather different code.

rcomer:
@larsbarring no I haven't made a branch for this.  I only got as far as playing with the units in a terminal to produce the output I showed.

larsbarring:
OK, I will give it a try.

Issue #216, opened 2022-02-25
Bug or feature: malformatted datetimes in time units go unnoticed


larsbarring:
## 🐛 Bug Report
By incidence I noticed that various "strange" datetimes are accepted:

```python
from cf_units import Unit
unit_str = [
    ("days since 1582-1010", "standard", "Error: generally malformed"),
    ("days since 1582-10-10", "standard", "Warning: date in Julian/Gregorian gap"),
    ("days since 1582-10-10", "julian", "Perfectly fine"),
    ("days since 1582-10-10", "proleptic_gregorian", "Perfectly fine"),
    ("days since haywire", "standard", "Captured error"),
    ("days since epoch", "standard", "Perfectly fine"),
    ("days since 0-0-0", "standard", "Error: wrong month/day"),
    ("days since 0-0-0", "julian", "Error: wrong month/day"),
    ("days since 0-0-0", "proleptic_gregorian", "Error: wrong month/day"),
    ("days since 0-1-1", "standard", "Warning: deprecated 'climatology' indicator"),
    ("days since 0-1-1", "julian", "Warning: deprecated 'climatology' indicator"),
    ("days since 0-1-1", "proleptic_gregorian", "Perfectly fine"),
    ("days since -1-1-1", "standard", "Error: negative year"),
    ("days since -1-1-1", "julian", "Error: negative year"),
    ("days since -1-1-1", "proleptic_gregorian", "Perfectly fine"),
]

for s in unit_str:
    try:
        u = Unit(s[0], calendar=s[1])
    except ValueError as e:
        u = str(e)
    print(f"({s[0]}, {s[1]}  ---> {u}   :   {s[2]}")

(days since 1582-1010, standard  ---> days since 1582-1010   :   Error: generally malformed
(days since 1582-10-10, standard  ---> days since 1582-10-10   :   Warning: date in Julian/Gregorian gap
(days since 1582-10-10, julian  ---> days since 1582-10-10   :   Perfectly fine
(days since 1582-10-10, proleptic_gregorian  ---> days since 1582-10-10   :   Perfectly fine
(days since haywire, standard  ---> [UT_SUCCESS] Failed to parse unit "days since haywire"   :   Captured error
(days since epoch, standard  ---> days since 1970-01-01 00:00:00   :   Perfectly fine
(days since 0-0-0, standard  ---> days since 0-0-0   :   Error: wrong month/day
[...]

PR #217, opened 2022-02-26
PoC fix issue 216

larsbarring:
### Description

Added "parser" to check validity of date/time part of time unit string. The "parser" is in fact a low level private function from `cftime`. Whether this is appropriate I am not sure, but implementing the same functionality more or less from scratch is beyond my capacity. [Hence the "PoC"]. 

From the added unit tests it *seems* -- in principle -- that the examples in #216 are handled. Only "in principle" because the actual errors generated by the cftime function are not properly captured. But again, understanding how cython exceptions interact with pytest is too advanced for me.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -20,6 +20,9 @@ from contextlib import contextmanager
 import cftime
 import numpy as np
 
+# low level cftime function for checking date part of unit string in Unit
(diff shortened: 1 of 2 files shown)


larsbarring:
Hrm, this was premature, naive and buggy ....

PR #236, opened 2022-04-27
fix ci failures

bjlittle:
### Description
This PR adds the `setuptools-scm` and `setuptools-scm-git-archive` as explicit package dependencies of `cf-units`, in order to resolve `AttributeError: module 'cf_units' has no attribute '__version__'` failures within CI e.g., see this recent [GHA run](https://github.com/SciTools/cf-units/runs/6156444892?check_suite_focus=true).

Apart from added the appropriate package dependencies, we also required to:
- bump the `cache-build` numbers
- include an update of the lock-files for the new dependencies


trexfeathers (approved):
Thanks @bjlittle! An important lesson learned

bjlittle:
Awesome, thanks @trexfeathers :beers:

Issue #202, opened 2021-10-27
Wrong dtype of return in date2num


zklaus:
## 📚 Documentation

The documentation for [`cf_units.Unit.date2num`](https://github.com/SciTools/cf-units/blob/a69d7c81cc8a0da4dada191a0861020bd56f33f0/cf_units/__init__.py#L1944) and [`cf_units.date2num`](https://github.com/SciTools/cf-units/blob/a69d7c81cc8a0da4dada191a0861020bd56f33f0/cf_units/__init__.py#L352) stipulates that both functions return a `float, or numpy.ndarray of floats`. However, since the return value comes directly from `cftime.date2num`, this will regularly return "an array of integers" since cftime version 1.4.0 (see [the changelog](https://github.com/Unidata/cftime/blob/b3e5c740106051e7dcff500ee6de979091f5a37c/Changelog#L47)).

pp-mo:
## Re-opened
I just realised that in addition to [`Unit.date2num(dates)`](https://github.com/SciTools/cf-units/blob/0ba82319a94b3d6b73ae8512507e04709869b4f9/cf_units/__init__.py#L1852), we **_also_** have a [`cf_units.date2num(dates, unit, calendar)`](https://github.com/SciTools/cf-units/blob/0ba82319a94b3d6b73ae8512507e04709869b4f9/cf_units/__init__.py#L329),
which is basically a different package for the same behaviours.

So unfortunately I think **_that_** docstring needs updating too.

**and...** we really need a whatsnew entry for these changes, too.
Sorry @ukmo-ccbunney I was so enthusiastic about #467 that I seem to have failed to review carefully !

ukmo-ccbunney:
Fixed by #483

PR #245, opened 2022-05-26
Purge unittest?

rcomer:
### Description

I learned of these tools to help convert unittest tests to pytest-style tests

* [unittest2pytest](https://github.com/pytest-dev/unittest2pytest)
* [pytestify](https://github.com/dannysepler/pytestify)

So obviously I wanted to see what happens if you point them at cf_units!  The `udunits` tests needed a little manual intervention, because the pattern

```python
with pytest.raises(SomeError) as thing:
```
gives you a different `thing` to 

```python
with self.assertRaises(SomeError) as thing:
```

Otherwise, this was remarkably painless.

The final commit just gets rid of the pattern `import cf_units as unit` from one test module.  I assume that was a hangover from when `cf_units` was taken out of the `iris.unit` module.

### DIFF
--- a/cf_units/tests/integration/test__num2date_to_nearest_second.py
+++ b/cf_units/tests/integration/test__num2date_to_nearest_second.py
@@ -6,7 +6,6 @@
 """Test function :func:`cf_units._num2date_to_nearest_second`."""
 
 import datetime
-import unittest
 
 import cftime
 import numpy as np
@@ -343,4 +342,4 @@ class Test:
 
 
 if __name__ == "__main__":
-    unittest.main()
+    pytest.main([__file__])
--- a/cf_units/tests/integration/test__num2date_to_nearest_second.py
+++ b/cf_units/tests/integration/test__num2date_to_nearest_second.py
@@ -339,7 +339,3 @@ class Test:
         ]  # rounded up to this
 
         self.check_timedelta(nums, units, expected)
-
-
-if __name__ == "__main__":
-    pytest.main([__file__])
--- a/cf_units/tests/integration/test__num2date_to_nearest_second.py
+++ b/cf_units/tests/integration/test__num2date_to_nearest_second.py
@@ -16,7 +16,7 @@ import cf_units
 from cf_units import _num2date_to_nearest_second
 
 
-class Test(unittest.TestCase):
+class Test:
     def setup_units(self, calendar):
         self.useconds = cf_units.Unit("seconds since 1970-01-01", calendar)
         self.uminutes = cf_units.Unit("minutes since 1970-01-01", calendar)
--- a/cf_units/tests/integration/test__num2date_to_nearest_second.py
+++ b/cf_units/tests/integration/test__num2date_to_nearest_second.py
@@ -10,6 +10,7 @@ import unittest
 
 import cftime
 import numpy as np
+import pytest
 
 import cf_units
 from cf_units import _num2date_to_nearest_second
@@ -27,8 +28,8 @@ class Test(unittest.TestCase):
             res = _num2date_to_nearest_second(
                 num, unit, only_use_cftime_datetimes=only_cftime
             )
-            self.assertEqual(exp, res)
-            self.assertIsInstance(res, type(exp))
+            assert exp == res
+            assert isinstance(res, type(exp))
 
     def check_timedelta(self, nums, units, expected):
         for num, unit, exp in zip(nums, units, expected):
@@ -36,15 +37,15 @@ class Test(unittest.TestCase):
             res = _num2date_to_nearest_second(num, unit)
             delta = res - epoch
             seconds = np.round(delta.seconds + (delta.microseconds / 1000000))
-            self.assertEqual((delta.days, seconds), exp)
+            assert (delta.days, seconds) == exp
 
     def test_scalar(self):
         unit = cf_units.Unit("seconds since 1970-01-01", "gregorian")
         num = 5.0
         exp = datetime.datetime(1970, 1, 1, 0, 0, 5)
         res = _num2date_to_nearest_second(num, unit)
-        self.assertEqual(exp, res)
-        self.assertIsInstance(res, cftime.datetime)
+        assert exp == res
+        assert isinstance(res, cftime.datetime)
 
     def test_sequence(self):
         unit = cf_units.Unit("seconds since 1970-01-01", "gregorian")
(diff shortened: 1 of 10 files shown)


cf_units/tests/integration/test__num2date_to_nearest_second.py
```diff
@@ -342,4 +342,4 @@ def test_fractional_second_365_day(self):
 
 
 if __name__ == "__main__":
-    unittest.main()
+    pytest.main([__file__])
```
rcomer: This retains the ability to run individual test modules as

`python test_stuff.py`

May now be redundant as you can just point `pytest` at a single test module.  I note that neither Matplotlib nor Cartopy have `ifmain`s in their test modules.

pp-mo: I would say we don't really have a need for this style anymore.
PytTest code does not really follow "normal" Python code idioms, due to its extensive use of inspection and patching. 
(even unittest is a bit like that, in discovery and setup/teardown)
So I think in adopting it we can simply accept that you will _always_ run it via pytest.

rcomer: Now removed https://github.com/SciTools/cf-units/pull/245/commits/d0bb08947dd6c919f796cde8ca89a1346fd0ea72

cf_units/tests/integration/test_date2num.py
```diff
@@ -24,7 +24,7 @@ def test_single(self):
         res = date2num(date, self.unit, self.calendar)
         # num2date won't return an exact value representing the date,
         # even if one exists
-        self.assertAlmostEqual(exp, res, places=4)
```
pp-mo, line 27: Given the [definition of assertAlmostEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual),
Especially "Note that these methods round the values to the given number of decimal places (i.e. like the [round()](https://docs.python.org/3/library/functions.html#round) function) and not significant digits." ...

I wonder if it would be better/safer to translate these differently.
So, I think something like ...
`assert round(exp, 7) == round(res, 7)`
should be more equivalent to the older form ?

In my experience with PyTest, it's also _always_ worth asking "what would a test failure message look like", since this is generally less nuanced than the older 'assertXXX' methods.
So, in this case e.g. "123.010426 != 123.010433" is going to be more helpful than "1.2e-6 != 0.0"

cf_units/tests/unit/test__udunits2.py
```diff
@@ -110,98 +109,98 @@ def setUp(self):
...
         _ud.get_converter(self.metre, self.yard)
 
     def test_get_converter_invalid(self):
-        with self.assertRaises(_ud.UdunitsError) as cm:
+        with pytest.raises(_ud.UdunitsError) as ex:
             _ud.get_converter(self.metre, self.second)
-        ex = cm.exception
 
-        self.assertEqual(ex.status_msg(), "UT_MEANINGLESS")
+        assert ex.match("UT_MEANINGLESS")
```
pp-mo: As above, can replace 
```
with ... as ex: 
    ... 
assert ex.match(msg)
```
with 
````
with raises(..., match=msg):
   ...
`

File: cf_units/tests/unit/test__udunits2.py [hunk 4]
```diff
@@ -110,98 +109,98 @@ def setUp(self):
...
+        assert time_since is not None
 
+    @pytest.mark.xfail
     def test_offset_by_time_invalid(self):
-        with self.assertRaises(_ud.UdunitsError) as cm:
+        with pytest.raises(_ud.UdunitsError) as ex:
             _ud.offset_by_time(self.metre, -31622400.0)
-        cm.exception
+
+        assert ex.match("UT_MEANINGLESS")
```
pp-mo: As above, would this be better as a `raises(errtype, match=msg)` ?

cf_units/tests/unit/test__udunits2.py
```diff
@@ -32,24 +32,23 @@ def test_read_xml(self):
...
 
     def test_read_xml_invalid_path(self):
-        with self.assertRaises(_ud.UdunitsError) as cm:
+        with pytest.raises(_ud.UdunitsError) as ex:
             _ud.read_xml(b"/not/a/path.xml")
-        ex = cm.exception
 
-        self.assertEqual(ex.status_msg(), "UT_OPEN_ARG")
-        self.assertEqual(ex.errnum, errno.ENOENT)
+        assert ex.match("UT_OPEN_ARG")
```
pp-mo: There are a few of these ...
I think it would be better to replace the `with raises(errtype) as ex: ... ex.match(msg)` with the simpler
`with raises(errtpye, match=msg):`

pp-mo (changes requested):
Respect for wading through all of this !
Some tiny suggestions, but basically all good !

cf_units/tests/integration/test_date2num.py
rcomer: Now I think about it (rather than just blindly following what `unittest2pytest` gave me) these days `date2num` will give you an `int` if it can - see #202.  I checked and both these tests are indeed getting `numpy.int64` types out.  So seems safe to just use equality.

rcomer:
Thanks @pp-mo, I've addressed your comments.  I'm at a loss what to do about Codacy's objection though.

cf_units/tests/integration/test_date2num.py
rcomer: Note: if I had used `pytestify` first, it would have changed it to `assert a == pytest.approx(b)`.

```
        a = 1
        b = 2
>       assert a == pytest.approx(b)
E       assert 1 == 2 ± 2.0e-06
E         comparison failed
E         Obtained: 1
E         Expected: 2 ± 2.0e-06
```

pp-mo: 😮  Wow.  I wasn't kidding when I said that PyTest was somewhat unlike normal Python code.
I guess that approx(x) must be a thing with a rather clever comparison behaviour.

pp-mo:
>  @rcomer I'm at a loss what to do about Codacy's objection though.

I think it is because 'self' is unused in this method.
So I suppose it "wants" you to remove it from the signature + add "@staticmethod"
But I don't know whether PyTest would take kindly to that or not.  
If not, you can just add a line with "self" on it, to evaluate but (presumably) do nothing.

rcomer:
I tried removing that method entirely, and I couldn't see any change in the test output.  🤷‍♀️

pp-mo (approved):

pp-mo:
Thanks @rcomer.
Not fun work, but very needed !

PR #271, opened 2022-07-11
fix rtd

bjlittle:
### Description
This PR reboots the building of documentation on `readthedocs`.

It also requires the following [user-defined environment variables](https://docs.readthedocs.io/en/stable/environment-variables.html#user-defined-environment-variables) to be configured on the `cf-units` RTD account:
- `UDUNITS2_INCDIR`
- `UDUNITS2_LIBDIR`
- `UNUNITS2_XML_PATH`

These environment variables are required as RTD performs a `pip install` and as a result attempts to build the binary wheel from the sdist of the code.

See the successfully [rendered documentation](https://bjlittle-cf-units.readthedocs.io/en/latest/) associated with this PR on my private clone of the `cf-units` RTD.

PR #272, opened 2022-07-12
rtd remove unshallow

### LINKED ISSUE #267: RTD is failing to build
## 🐛 Bug Report

For quite a while now the `cf-units` documentation on `readthedocs` has been failing to build and publish successfully, see [here](https://readthedocs.org/projects/cf-units/builds/).

bjlittle:
### Description
This PR resolves this issue on RTD building for `main`:

image

Which didn't present itself whilst testing on a private branch for #271.

[Here](https://readthedocs.org/projects/bjlittle-cf-units/builds/17416070/) is the successful build of the documentation for this pull-request.

Closes #267

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,7 +6,6 @@ build:
     python: mambaforge-4.10
   jobs:
     post_checkout:
-      - git fetch --unshallow
       - git fetch --all
 
 conda:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,6 +6,9 @@ build:
     python: mambaforge-4.10
   jobs:
     post_checkout:
+      # reference: https://github.com/SciTools/cf-units/pull/272
+      # perform a "git fetch --unshallow" when the cf-units repo
+      # becomes incomplete i.e., there is a .git/shallow.
       - git fetch --all
 
 conda:


bjlittle:
@trexfeathers I can't quite explain this... don't quite understand why this issue didn't happen for #271 when testing.

trexfeathers:
> @trexfeathers I can't quite explain this... don't quite understand why this issue didn't happen for #271 when testing.

@bjlittle I also want to understand why it's not a problem for Iris - see [`iris/.readthedocs.yml#L8-10`](https://github.com/SciTools/iris/blob/2cbaec747820410cce41db004a41e68d6ff63ee8/.readthedocs.yml#L8-L10)

trexfeathers:
This could plausibly be down to a difference in config within RTD. I set up cf-units as bare-bones as I could, so perhaps Iris has a better config that we should be mimicking.

bjlittle:
@trexfeathers I think it's a difference in the repos. `iris` has more history.

Also, it's clear that you do need to do `git fetch --unshallow` for `iris` otherwise the version is wrong.

.readthedocs.yml
```diff
@@ -6,7 +6,6 @@ build:
     python: mambaforge-4.10
   jobs:
     post_checkout:
-      - git fetch --unshallow
```
trexfeathers, line 9: @bjlittle and I now understand that `git fetch --unshallow` is invalid for cf-units, because cf-units (unlike Iris) is NOT a shallow repository. This might change in future if anything 'nasty' happens in the Git history, but the aim/hope is that it remains unshallow.

Please could we have an explanatory comment here, in case things change or this config is used as a template for another repo.

trexfeathers (changes requested):
@bjlittle and I now understand that `git fetch --unshallow` is invalid for cf-units, because cf-units (unlike Iris) is NOT a shallow repository. This might change in future if anything 'nasty' happens in the Git history, but the aim/hope is that it remains unshallow.

.readthedocs.yml
bjlittle: @trexfeathers Done :+1:

See 61d3280.

bjlittle: We could auto-detect `.git/shallow`, but perhaps that is a tad over-engineering.

I'm erring on the side of not doing that here.

trexfeathers (approved):

PR #179, opened 2021-06-21
Linting

trexfeathers:
### Description

Closes #171.

Adds the 'standard' SciTools linters, and tidies some other small things.

PR #273, opened 2022-07-12
update readme badges

bjlittle:
### Description
This PR updates and refreshes the `README.md` badges.

Notably, badges to the GHA CI and RTD have been added, and the badge to the now defunct `cirris-ci` has been removed.

The changes render as follows:

image

See [here](https://github.com/bjlittle/cf-units/tree/add-badges#----cf-units).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,16 @@
 [![Documentation Status](https://readthedocs.org/projects/cf-units/badge/?version=latest)](https://cf-units.readthedocs.io/en/latest/?badge=latest)
 [![Coverage Status](https://codecov.io/gh/SciTools/cf-units/branch/main/graph/badge.svg?token=6LlYlyTUZG)](https://codecov.io/gh/SciTools/cf-units)
 [![pre-commit.ci status](https://results.pre-commit.ci/badge/github/SciTools/cf-units/main.svg)](https://results.pre-commit.ci/latest/github/SciTools/cf-units/main)
-,
+\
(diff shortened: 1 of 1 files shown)


trexfeathers (changes requested):
You're not gonna like this 😅

I'm quite strongly against using HTML in markdown files unless it's absolutely necessary. It's not consistent with the `​.md​` format, and is harder to read.

I appreciate that this is the only way to get centralised text/badges, but I just don't think it's worth it. That's why I introduced markdown badges in #179 a year ago (which you were happy to merge).

It would also obvs be way easier for me to review this if the format was consistent.

bjlittle:
I personally don't mind HTML within Markdown, particularly when Markdown simply can't do the job.

So now we have this instead...

image

😞

trexfeathers:
> I personally don't mind HTML within Markdown, particularly when Markdown simply can't do the job.
> 
> So now we have this instead...
> 
> 😞

Sounds like we both have strong opinions. I'll happily bow to a third opinion from another dev (which would make it a quorum either way).

bjlittle:
@trexfeathers I honestly don't greatly mind. Tis fine :+1:

trexfeathers (changes requested):
Badges look good to me, and feel free to try alternatives to the commas.

Codacy has also picked up some boo boos I made in my original reformatting. Could you address those?

bjlittle:
So we now have this...
image

trexfeathers (approved):

PR #279, opened 2022-07-21
Remove in-house long time interval checking.

### LINKED ISSUE #274: Align long time interval check with cftime
We currently blanket disallow `month` or `year` intervals when passing operations down to cftime:

https://github.com/SciTools/cf-units/blob/37c840008a12e6f8495ebf400c7c7839cab173f1/cf_units/__init__.py#L1932-L1939

cftime disallows these intervals, presumably because it requires consistent-length time increments. But cftime is more intelligent about this than cf-units is, e.g. `month` intervals are allowed for `360_day` calendars (Unidata/cftime#75).

We should make sure our own checks don't cover over details like the `360_day` exception. The Pythonic choice would be to do no extra checks at all in cf-units, simply relying on cftime to raise the appropriate error; is there a good reason why we don't do this?

trexfeathers:
### Description

Closes #274. Removes cf-units' own checking for long time intervals that might prevent certain operations.

This was introduced in #72, but cftime (previously netcdftime) continues to improve handling of long time intervals, and has a more informative error when they can't be handled. This makes our own checking superfluous and also overzealous.

I have removed all tests concerning long time interval errors, since it would be inappropriate for our tests to depend on the behaviour of cftime.

If this change is accepted: a sister PR on Iris will also be needed, since Iris includes some defensive coding to avoid failure when printing an invalid time `Unit` - [`iris.coords#L346`](https://github.com/SciTools/iris/blob/56a97d66b8404f96ef61fe2d52433eb58373f338/lib/iris/coords.py#L346). To allow for cftime's continuing improvements - including tolerance for previously invalid time spacings - it would be safest to use try-except during the printing rather than a more specific check.

<details><summary>Demonstration</summary>

```python
my_unit = Unit("months since 2010-01-01", calendar="standard")
print(my_unit.num2date(3))

```

```
  File "/tmp/persistent/repos/cf-units/tmp_long.py", line 4, in <module>
    print(my_unit.num2date(3))
  File "/tmp/persistent/repos/cf-units/cf_units/__init__.py", line 2043, in num2date
    return _num2date_to_nearest_second(
  File "/tmp/persistent/repos/cf-units/cf_units/__init__.py", line 564, in _num2date_to_nearest_second
    dates = cftime.num2date(time_values, **num2date_kwargs)
  File "src/cftime/_cftime.pyx", line 499, in cftime._cftime.num2date
  File "src/cftime/_cftime.pyx", line 99, in cftime._cftime._dateparse
ValueError: 'months since' units only allowed for '360_day' calendar

```

---

```python
my_unit = Unit("months since 2010-01-01", calendar="360_day")
print(my_unit.num2date(3))

```

```
2010-04-01 00:00:00
```

</details>


cf_units/__init__.py
```diff
@@ -964,36 +964,6 @@ def is_time_reference(self):
         """
         return self.calendar is not None
 
-    def is_long_time_interval(self):
```
pp-mo: All looks good except : for some backwards compatibility, it might be nice to deprecate this function instead of simply removing it ?

Checked : iris is currently using, as noted in the PR header notes.  
So that at least might prompt us to retain it for a grace period.

pp-mo (dismissed):
Otherwise all looks good

cf_units/__init__.py
trexfeathers: 947fbca630e6b421b443322bc68771e968f69733

bjlittle:
@trexfeathers ping (if you get time) 😉

trexfeathers:
@pp-mo I forgot about this, but it is now ready to merge. It would also fix at least 1 Iris issue. Could you take a quick look? Thanks!

cf_units/__init__.py
```diff
@@ -873,6 +874,12 @@ def is_long_time_interval(self):
         discrepancy means we cannot run self.num2date() on a time unit with
         a long time interval.
 
+        .. deprecated:: 3.2.0
```
bjlittle:
```suggestion
        .. deprecated:: 3.3.0
```

bjlittle (changes requested):
@trexfeathers Awesome, thanks! 😄 

Just a minor tweak and we're good to go 👍

cf_units/__init__.py
trexfeathers: Hilarious.

Thanks for spotting!

bjlittle (approved):

trexfeathers:
Thanks for getting this over the line @bjlittle!

PR #316, opened 2022-11-08
Deprecate `cf_units.util.approx_equal`

rcomer:
### Description

As for https://github.com/SciTools/iris/pull/4514, I don't think we need to be maintaining this function.  Unlike the Iris version, I didn't bother updating the docstring, because this function isn't included in the cf-units docs!

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -15,6 +15,7 @@ See also: `UDUNITS-2
 """
 
 import copy
+import math
 from contextlib import contextmanager
 
 import cftime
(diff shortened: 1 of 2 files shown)


lbdreyer (commented):
Thanks for doing this @rcomer !

The trouble with not including the deprecation in the docstring is that there is always the risk that we change the cf-units docs such that it starts publishing the cf_units.util API, and that this is done before we remove cf_units.util.approx_equal. Do you mind adding a small mention of the deprecation on the off chance this happens?

rcomer:
@lbdreyer done.  Though obviously I have no way to check whether it would render as expected!

lbdreyer (approved):

rcomer:
Thanks @lbdreyer!

PR #331, opened 2023-01-05
Relicense from LGPL-3 to BSD-3

lbdreyer:
All the code in this repo comes from contributors who have now either signed the new CLA or are employees of the Met Office.

The new CLA enables us to relicense to BSD-3.

This PR:
* adds a BSD-3 LICENSE file (to replace the old COPYING* files)
* Updates the pypi classifiers/metadata etc
* Updates the file license headers


bjlittle:
@lbdreyer Test failure to investigate, but not your doing... 🤔

cf_units/tests/__init__.py
```diff
@@ -1,5 +1,4 @@
 # Copyright cf-units contributors
-#
-# This file is part of cf-units and is released under the LGPL license.
-# See COPYING and COPYING.LESSER in the root of the repository for full
-# licensing details.
+
```
pp-mo: I think, according to the license-headers test, this line should have a "#", not just blank.
```suggestion
#
```

rcomer:
I think that test failure is just a method that shouldn't start with "test".

https://github.com/SciTools/cf-units/blob/2a164dc3d3333bfce5b6855946b3be25e0dab079/cf_units/tests/unit/unit/test_Unit.py#L82-L94

Edit: nope, that's just a warning I'm looking at.  I'll open a PR to fix the warning anyway....

cf_units/tests/integration/test_num2date.py
```diff
@@ -1,8 +1,7 @@
 # Copyright cf-units contributors
 #
-# This file is part of cf-units and is released under the LGPL license.
-# See COPYING and COPYING.LESSER in the root of the repository for full
-# licensing details.
+# This file is part of cf-units and is released under the BSD license.
+# See LICENSE in the root of the repository for full licensi
```
pp-mo: Bit went missing here.
```suggestion
# See LICENSE in the root of the repository for full licensing details.
```

pp-mo (changes requested):
> @bjlittle not your doing

Not so sure.
 
I think the two files mentioned **_did_** get missed in updating the licence header.
So, I think these tweaks may fix what was wrong?
Haven't really checked anything else yet, but I'm sure it's right in spirit.

lbdreyer:
Thanks for catching the problem @pp-mo !


The changes should now be correct

bjlittle (approved):
👍

pp-mo (approved):
Re-reviewing this to approve.
But I can't understand why this is necessary, since it lists no unresolved conversations.
Seen this before, I think it is possibly a GitHub bug ??

lbdreyer:
Thanks @pp-mo !

PR #332, opened 2023-01-06
setup method is not a test

rcomer:
### Description

Pytest was throwing a warning because it ran this method as a test, and tests shouldn't return anything.  Changing the name ensures pytest doesn't think it's a test.


lbdreyer (approved):
Thanks @rcomer ! Good to get this fixed since the pytest warning states that it will become a error in future pytest versions

PR #348, opened 2023-04-17
support py311 locks

bjlittle:
### Description

This PR extends the conda lock file generation to support `py311`.

After this lands, I'll extend the `tox` testing to cover `py311` too, but in a separate PR, as the lock files require to be available before we turn on testing.

We can manually run the `ci-locks` to auto generated the PR containing `py311` dependencies.

### DIFF
--- a/.github/workflows/ci-locks.yml
+++ b/.github/workflows/ci-locks.yml
@@ -27,7 +27,7 @@ jobs:
       ENV_NAME: "ci-locks"
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


bjlittle:
The fact that `cf-units` 3.2 is built as a binary wheel for `py311` is okay AFAIC.

I don't think there is a need to do a post release to include this test coverage, but others may disagree. Anyways, I'm happy to do that if you guys wanted to :+1:

rcomer (approved):

PR #349, opened 2023-04-17
update deps min pins

bjlittle:
### Description

This PR bumps some minimum dependency pins for `cftime`, `setuptools` and `setuptools-scm`.

Note that, the `cftime` minimum pin here aligns with the minimum pin within the `cf_units-feedstock` recipe, see [here](https://github.com/conda-forge/cf_units-feedstock/blob/main/recipe/meta.yaml#L34).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,8 @@
 [build-system]
 # Defined by PEP 518
 requires = [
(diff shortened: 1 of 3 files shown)


pelson (approved):
LGTM

PR #351, opened 2023-04-17
remove CHANGES file

bjlittle:
### Description

This PR purges the `CHANGES` file, which IMHO is simply a redundant historical file.

If anyone is really interested in it, then it's captured in our GitHub repo history.

Note that, if we merge this after #350, then we should remove the `exclude CHANGES` entry from the `MANIFEST.in`... or do the same in #350 if this PR is merged first.


rcomer (approved):
Looks like these are changes from before cf_units was moved out of Iris, so could be more awkward to find (particularly if you didn't know cf_units used to be part of Iris).  OTOH I'd never seen this file before, and didn't think to look here when I was previously trying to find old `Unit` changes in Iris.  I also don't think this file is comprehensive.

So I give you an approval and you can self-merge if you like.

PR #352, opened 2023-04-18
reorganise readme badges

bjlittle:
### Description

This PR reorganises the `README.md` badges into the following:

image


pp-mo (changes requested):
Looks cool, they are much less bewildering when grouped like that.

Some possible additions, and is it finicky to suggest different orderings ?
After some thought, changes I think I might make:
  * [ ] move 'ci-tests' to front of CI section, as I think its the most important entry
  * [ ] sections order : put "community" under "repo", and "package" last (as it's effectively a download link reference).
  * [ ] actually I'm not sure about 'Community'.  If we had a code-of-conduct badge it could go there, but the DOI as we have it is version-specific : so _that_ could just go in 'package'.
  * [ ] didn't we have a badge for "platform: agnostic" package type? -- would go in 'package' section (not in Meta)
  * [ ] include "imports: isort" in the Meta section -- this one seems to have disappeared, was there a reason for that ?

I'm a bit puzzled because I think it may have changed under my feet while I was writing some notes, but I can't quite see how !

bjlittle:
@pp-mo 
>     * [ ]  move 'ci-tests' to front of CI section, as I think its the most important entry
> 
It is, but I've put them in alphabetical order, otherwise they're just random.

I agree `ci-tests` are important, but to be fair, they're not hard to find as-is.

>     * [ ]  sections order : put "community" under "repo", and "package" last (as it's effectively a download link reference).
> 

The sections are ordered alphabetically, otherwise it's a bit random.

Once we agree on the [CoC poll](https://github.com/SciTools/iris/discussions/5197) then the badge for that will go in the `💬 Community` section, so I think it has relevance. Also, we could also add a link to the `cf-units` GH Discussions... infact I can do that one now.

>     * [ ]  actually I'm not sure about 'Community'.  If we had a code-of-conduct badge it could go there, but the DOI as we have it is version-specific : so _that_ could just go in 'package'.

For me, the community reference the DOI to cite us. That's my thinking on it, but happy to move it to the package :+1: 

> 
>     * [ ]  didn't we have a badge for "platform: agnostic" package type? -- would go in 'package' section (not in Meta)

`cf-units` isn't pure python, so isn't platform agnostic. It really does care about the platform and the chipset as we've got cythonized code wrappend around a C-library. So, I'm not sure what you mean :thinking: 

> 
>     * [ ]  include "imports: isort" in the Meta section -- this one seems to have disappeared, was there a reason for that ?

I'll re-add those :+1: 
> 
> 
> I'm a bit puzzled because I think it may have changed under my feet while I was writing some notes, but I can't quite see how !

pp-mo:
> `cf-units` isn't pure python

Sorry, don't quite know what I was thinking there.

pp-mo (approved):

pp-mo:
Thanks @bjlittle , clearly an improvement !

bjlittle:
@pp-mo Awesome, thanks :rocket:

bjlittle:
@pp-mo when #350 lands then the `ci-status` badge will come to life (hopefully)

image

PR #353, opened 2023-04-18
adopt ruff

bjlittle:
### Description

This PR adopts [ruff](https://github.com/charliermarsh/ruff) and drops the explicit use of `isort` and `flake8`.

Running `ruff` across the codebase rediscovers a whole collections of exceptions that clearly have been missed for `flake8` et al compliance.

- [ ] Once #352 is merged, then I'll rebase and add the `ruff` badge.

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -13,9 +13,9 @@ See also: `UDUNITS-2
 
 """
 
+from contextlib import contextmanager
 import copy
 import math
-from contextlib import contextmanager
 
 import cftime
 import numpy as np
@@ -552,15 +552,13 @@ def _num2date_to_nearest_second(
     # later versions, if one wished to do so for the sake of consistency.
     cftime_unit = unit.cftime_unit
     time_units = cftime_unit.split(" ")[0]
-    has_half_seconds = np.logical_and(
-        time_units == "seconds", time_values % 1.0 == 0.5
-    )
-    num2date_kwargs = dict(
-        units=cftime_unit,
-        calendar=unit.calendar,
-        only_use_cftime_datetimes=only_use_cftime_datetimes,
-        only_use_python_datetimes=only_use_python_datetimes,
-    )
+    has_half_seconds = np.logical_and(time_units == "seconds", time_values % 1.0 == 0.5)
+    num2date_kwargs = {
+        "units": cftime_unit,
+        "calendar": unit.calendar,
+        "only_use_cftime_datetimes": only_use_cftime_datetimes,
+        "only_use_python_datetimes": only_use_python_datetimes,
+    }
     dates = cftime.num2date(time_values, **num2date_kwargs)
     try:
         # We can assume all or none of the dates have a microsecond attribute
@@ -664,11 +662,9 @@ def _ud_value_error(ud_err, message):
 
     ud_msg = ud_err.error_msg()
     if ud_msg:
-        message = "{}: {}".format(message, ud_msg)
+        message = f"{message}: {ud_msg}"
 
-    message = "[{status}] {message}".format(
-        status=ud_err.status_msg(), message=message
-    )
+    message = f"[{ud_err.status_msg()}] {message}"
 
     return ValueError(message)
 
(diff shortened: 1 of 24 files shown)


rcomer:
Lots of style changes here.  Is it worth adding a `.git-blame-ignore-revs`?

cf_units/__init__.py
```diff
@@ -552,15 +552,13 @@ def _num2date_to_nearest_second(
...
-    has_half_seconds = np.logical_and(
-        time_units == "seconds", time_values % 1.0 == 0.5
-    )
-    num2date_kwargs = dict(
-        units=cftime_unit,
-        calendar=unit.calendar,
-        only_use_cftime_datetimes=only_use_cftime_datetimes,
-        only_use_python_datetimes=only_use_python_datetimes,
-    )
+    has_half_seconds = np.logical_and(time_units == "seconds", time_values % 1.0 == 0.5)
```
rcomer, line 555: Seems like ruff is overriding black here.  Maybe we shouldn't use both?

https://github.com/SciTools/cf-units/blob/cf4d66d002dc1028d33c7d80a66396a1856de6f4/pyproject.toml#L48-L49

cf_units/__init__.py
```diff
@@ -552,15 +552,13 @@ def _num2date_to_nearest_second(
...
+    num2date_kwargs = {
+        "units": cftime_unit,
+        "calendar": unit.calendar,
+        "only_use_cftime_datetimes": only_use_cftime_datetimes,
+        "only_use_python_datetimes": only_use_python_datetimes,
+    }
```
rcomer, line 561: I'm curious about why this is the preferred form for specifying a dictionary.

cf_units/__init__.py
rcomer: Actually it looks like a lot of the changes in this PR are just increasing the allowed line length.  Maybe it would be better to tell ruff to keep to 79 characters.  I've no objection in general to an 88 character limit, but it might be good to separate concerns of tool choices vs style choices.  Line limit could be increased as a follow-up PR.

trexfeathers:
I really didn't want to get involved here but based on @rcomer's comments I think we need a team discussion. I think we can all agree that we don't want inconsistency between our repos, so adopting ruff for cf-units would therefore create an implicit goal to use it for all our repos, and that's a big decision.

cf_units/__init__.py
```diff
@@ -1881,25 +1850,20 @@ def convert(self, value, other, ctype=FLOAT64, inplace=False):
...
                     # Strict type check of numpy array.
                     if result.dtype.type not in (np.float32, np.float64):
-                        raise TypeError(
-                            "Expect a numpy array of '%s' or '%s'"
-                            % np.float32,
-                            np.float64,
+                        emsg = (
+                            f"Expect a numpy array of '{np.float32}' or '{np.float64}'"
                         )
+                        raise TypeError(emsg)
```
rcomer, line 1856: This refactor seems more clever than black 🤔

cf_units/__init__.py
rcomer: Looks like you just need a couple of lines in the `pyproject.toml` to reduce the character limit

```
[tool.ruff]

line-length = 79
```
I'd be interested to see how much smaller that makes the diff, and it might help inform the discussion at https://github.com/SciTools/iris/discussions/5254.  Looks like a lot of the other changes are introducing f-strings to replace old-style string formatting, which seems fair enough.

pp-mo: > Seems like ruff is overriding black here. Maybe we shouldn't use both?
> 
> https://github.com/SciTools/cf-units/blob/cf4d66d002dc1028d33c7d80a66396a1856de6f4/pyproject.toml#L48-L49

I've have wondered before whether we are using overlapping tools.  I think there is scope for black to disagree with flake8 ?

rcomer: Ruff docs say they are compatible with black _provided_ the line lengths are set the same.
https://beta.ruff.rs/docs/faq/#is-ruff-compatible-with-black

cf_units/__init__.py
rcomer: Huh.  Maybe this is why

```python
In [5]: %timeit x = {"a":1, "b":2}
70.2 ns ± 1.32 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit x = dict(a=1, b=2)
119 ns ± 2.61 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

```

PR #357, opened 2023-04-21
refresh supported python distros

bjlittle:
### Description

This PR drops our support for `py38` and extends to `py311`, as per the [NEP-29 Drop Schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule).


rcomer (approved):
LGTM

PR #368, opened 2023-06-27
Update antlr version to 4.11.1

rcomer:
### Description

Helps with #313.  Not sure if it closes it, as it might still be good to make the dependency optional.

I figured out what to do with a combination of [the README](https://github.com/SciTools/cf-units/blob/main/cf_units/_udunits2_parser/README.md) and [this helpful comment and the PR it links to](https://github.com/antlr/antlr4/issues/3997#issuecomment-1363960317).  The only changes I made manually were the version number in three files.  All the other changes are from running the `compile.py` script.  I have not attempted to understand them!

The newest version of this dependency available from conda-forge is 4.13.0, but it was noted at https://github.com/SciTools/cf-units/issues/313#issuecomment-1601899713 that lots of other packages use 4.11.1, so it is likely helpful to be consistent.

Tests pass for me locally.  ~I do not expect them to pass in CI because I have not updated the lockfiles.  Can anyone tell me how to do that locally?~


rcomer:
~Huh.  Tests do pass.  I guess `antlr4-python3-runtime` is getting installed independently of the lockfiles?~

cf_units/_udunits2_parser/parser/udunits2Parser.py
```diff
@@ -1,86 +1,1311 @@
...
+        90,
+        96,
+        101,
+        105,
+        108,
+        113,
+        116,
+        118,
+        123,
+        127,
```
rcomer, line 1238: Perhaps `black` should be skipped for these files 🤔

rcomer:
Added some notes for next time.

rcomer:
~Codacy seems to be stuck 😕~

rcomer:
I got a full set of green ticks! 🥳

trexfeathers (approved):
Taking over as @pp-mo has been deployed elsewhere for now.

Looks good to me, thanks for having a go @rcomer! ~~There might be some fallout with dependencies but I think we can fix post-merge.~~

EDIT: forgot I had the 'update branch' button, so that should hopefully address the dependency thing.

rcomer:
Thanks @trexfeathers 🥳

Issue #391, opened 2023-12-02
Support meson-python or similar


monego:
## 📰 Custom Issue

Extensions in Cython and other compiled languages can be built by newer build backends based on [PEP 517](https://learn.scientific-python.org/development/guides/packaging-compiled/) and [PEP 621](https://learn.scientific-python.org/development/guides/packaging-simple/). I expect this move to ease the side of packagers and others who have to build from source, as well as that of maintainers who would have less code to maintain than that of setup.py. The downside is that it adds an extra dependency.

I had this idea because I always get a circular import error while trying to package (building from source) cf-units for GNU Guix:

```
E   ImportError: cannot import name '_udunits2' from partially initialized module 'cf_units' (most likely due to a circular import) (/tmp/guix-build-python-cf-units-3.2.0.drv-0/cf-units-3.2.0/cf_units/__init__.py)
```

In 3.1.0 and before that error would happen in build time, before I could compile the Cython extensions. But in 3.2.0 it only happens when running the test suite. That hints of a change in setup.py, but I haven't investigated much.

ESadek-MO:
@SciTools/peloton don't really have the time or resources to complete this, so closing this as unplanned. Sorry for the inconvenience!

pp-mo:
Time to  reconsider this, I think ...

PR #392, opened 2023-12-07
Port from setuptools.build_meta to meson-python

monego:
### Description
This is an initial attempt at #391. [meson-python](https://meson-python.readthedocs.io/en/latest/) is usually adopted by large projects and may be overkill here, but cf-units is a good starting point to learn more about it. There is a Cython extension for udunits and an external project (antlr4) to manage (not yet implemented in this PR).

This change moves from setup.py and setup.cfg to Meson scripts and meson-python. Meson detects the udunits installation and adjusts the necessary paths accordingly. I'm also considering a fix for #313 with a Meson build option.

This PR is currently tested on Linux (with Debian 12 packages) only.

I added [Spin](https://github.com/scientific-python/spin) to run the build and test commands. All tests are passing (and one skipped and one xfail).

```bash
$ spin build
$ spin test
```

I installed all the other Python dependencies (listed in pyproject.toml) manually from pip. udunits is from the Debian repo.

What is missing:

- [ ] Restore `__version__`, check whether/how mesonpy can generate the version string from meson.build
- [ ] Check whether something important was left behind in setup.py and setup.cfg
- [ ] Check how this will affect the conda environment
- [ ] Adjust dependency version constraints
- [ ] Windows/Mac tests


trexfeathers:
From @SciTools/peloton: thanks so much for this @monego, we promise we're thinking about it - the main challenge is making time to look in more detail.

ESadek-MO:
@SciTools/peloton don't really have the time or resources to do complete this, so closing this as unplanned. Sorry for the inconvenience!

PR #396, opened 2023-12-15
Bump download/upload artifact gha version

bjlittle:
### Description

Require to bump the version of the `download-artifact` and `upload-artifact` to the latest version (`@3` -> `@4`) consistently within the same pull-request.

### DIFF
--- a/.github/workflows/ci-wheels.yml
+++ b/.github/workflows/ci-wheels.yml
@@ -72,7 +72,7 @@ jobs:
         UDUNITS2_XML_PATH: ${{ matrix.xml_path }}
         CIBW_ENVIRONMENT_PASS_LINUX: UDUNITS2_INCDIR UDUNITS2_LIBDIR UDUNITS2_XML_PATH
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
         name: pypi-artifacts
         path: ${{ github.workspace }}/wheelhouse/*.whl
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-wheels.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         UDUNITS2_XML_PATH: ${{ matrix.xml_path }}
         CIBW_ENVIRONMENT_PASS_LINUX: UDUNITS2_INCDIR UDUNITS2_LIBDIR UDUNITS2_XML_PATH
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
         name: pypi-artifacts
```
pp-mo, line 77: As we've seen the basic v3-to-v4 changes is that artifacts are now[ immutable and non-rewritable](https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes)
 
So I think the problem here is probably that the artifact name must be different for each usage, 
so, something like `pypi-artifacts-${{ matrix.os }}-${{ matrix.arch }}` might work ?

It seems like that, now we are storing different ones in different places, the other tasks will also probably need updating?  TBH I'm a bit out of my depth with this.

bjlittle: Yeah, there is now no artifact merging and names have to be unique

bjlittle:
Reference:

- https://github.com/actions/upload-artifact/tree/v4#not-uploading-to-the-same-artifact 😢 
- https://github.com/pypa/cibuildwheel/issues/1699
- https://github.com/actions/upload-artifact/issues/472

bjlittle:
I think the best option here for `cf-units` is **not** to migrate to `v4` for `upload-artifact` and consequently `download-artifact` due to the breaking change behaviour introduced.

I recommend that we sit tight and wait for the dust to settle, as community pressure may force a change or a documented best practice pattern may emerge for the common `cf-units` use case with `cibuildwheels` with a matrix of distros for wheels and sdist.

PR #436, opened 2024-07-17
Avoid issues with LC_NUMERIC locale

### LINKED ISSUE #435: Locale issue
## 🐛 Bug Report

`cf_units` fails to import when the `LC_NUMERIC` environmental variable is set to a locale that has a `,` instead of a `.` as the `decimal_point` character and Python sets the locale. This typically happens in [documentation builds with sphinx](https://github.com/search?q=repo%3Asphinx-doc%2Fsphinx%20setlocale&type=code). See https://github.com/pydata/xarray/issues/4257 for an example. I have encountered the same issue when building the [iris](https://github.com/SciTools/iris/) and [ESMValCore](https://github.com/ESMValGroup/ESMValCore/) documentation.

## How to Reproduce
Steps to reproduce the behaviour:

1. Run
`export LC_NUMERIC=nl_NL.UTF-8`
2. View locale settings by running `locale`:
```
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=nl_NL.UTF-8
LC_TIME=nl_NL.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=nl_NL.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=nl_NL.UTF-8
LC_NAME=nl_NL.UTF-8
LC_ADDRESS=nl_NL.UTF-8
LC_TELEPHONE=nl_NL.UTF-8
LC_MEASUREMENT=nl_NL.UTF-8
LC_IDENTIFICATION=nl_NL.UTF-8
LC_ALL=
```
3.
Run the following Python code to [set the locale](https://docs.python.org/3/library/locale.html#locale.setlocale)
```python
import locale
locale.setlocale(locale.LC_ALL, '')
import cf_units
```
4. Resulting stack trace:
```
  File "/home/bandela/src/scitools/cf_units/cf_units/__init__.py", line 187, in <module>
    _ud_system = _ud.read_xml()
                 ^^^^^^^^^^^^^^
  File "cf_units/_udunits2.pyx", line 194, in cf_units._udunits2.read_xml
    return wrap_system(csystem)
  File "cf_units/_udunits2.pyx", line 104, in cf_units._udunits2.wrap_system
    _raise_error()
  File "cf_units/_udunits2.pyx", line 184, in cf_units._udunits2._raise_error
    raise UdunitsError(status, errnum)
cf_units._udunits2.UdunitsError: UT_PARSE

During handling of the above exception, another exception occurred:

  File "/home/bandela/src/scitools/cf_units/cf_units/__init__.py", line 190, in <module>
    _ud_system = _ud.read_xml(config.get_xml_path())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cf_units/_udunits2.pyx", line 194, in cf_units._udunits2.read_xml
    return wrap_system(csystem)
  File "cf_units/_udunits2.pyx", line 104, in cf_units._udunits2.wrap_system
    _raise_error()
  File "cf_units/_udunits2.pyx", line 184, in cf_units._udunits2._raise_error
    raise UdunitsError(status, errnum)
cf_units._udunits2.UdunitsError: UT_PARSE

During handling of the above exception, another exception occurred:

  File "/home/bandela/src/scitools/cf_units/test.py", line 4, in <module>
    import cf_units
  File "/home/bandela/src/scitools/cf_units/cf_units/__init__.py", line 193, in <module>
    raise OSError(
OSError: [UT_PARSE] Failed to open UDUNITS-2 XML unit database
```

## Expected Behaviour

A successful import.

## Environment 
 - OS & Version: Ubuntu 24.04 LTS
 - cf-units Version: 3.2.1.dev57

## Additional Context
When I comment out the `suppress_warnings` context manager here https://github.com/SciTools/cf-units/blob/f57a7f1fdba4e9424e4dbbc01b614e521a88086c/cf_units/__init__.py#L182
I see the following error message:
```
Invalid numeric prefix value ".1"
parsing aborted
File "/home/bandela/mambaforge/envs/cf-units-dev/share/udunits/udunits2-prefixes.xml", line 43, column 25
parsing aborted
File "/home/bandela/mambaforge/envs/cf-units-dev/share/udunits/udunits2.xml", line 11, column 42
  File "/home/bandela/src/scitools/cf_units/test.py", line 4, in <module>
    import cf_units
  File "/home/bandela/src/scitools/cf_units/cf_units/__init__.py", line 187, in <module>
    _ud_system = _ud.read_xml()
                 ^^^^^^^^^^^^^^
  File "cf_units/_udunits2.pyx", line 194, in cf_units._udunits2.read_xml
    return wrap_system(csystem)
  File "cf_units/_udunits2.pyx", line 104, in cf_units._udunits2.wrap_system
    _raise_error()
  File "cf_units/_udunits2.pyx", line 184, in cf_units._udunits2._raise_error
    raise UdunitsError(status, errnum)
cf_units._udunits2.UdunitsError: UT_PARSE
```
some [searching trough the udunits2 code](https://github.com/search?q=repo%3AUnidata%2FUDUNITS-2+%22Invalid+numeric+prefix+value%22&type=code) suggests that it originates from using the [`strtod`](https://en.cppreference.com/w/c/string/byte/strtof) function in [this code](https://github.com/Unidata/UDUNITS-2/blob/c83da987387db1174cd2266b73dd5dd556f4476b/lib/xml.c#L1731-L1742), which uses the `decimal-point` locale setting.

bouweandela:
## 🚀 Pull Request

Avoid a crash when importing the library as described in #435.

Closes #435.


bouweandela:
The CI failures appear unrelated to this pull request.

pp-mo:
@bjlittle can you spare this some attention?  Looks simple enough to me, but I'm about to be unavailable for a couple of weeks..

cf_units/__init__.py
```diff
@@ -175,11 +177,25 @@ def suppress_errors():
...
 
 
+LOCALE_LOCK = threading.Lock()
+
+
+@contextmanager
+def us_locale():
+    with LOCALE_LOCK:
+        lc_numeric = locale.getlocale(locale.LC_NUMERIC)
+        locale.setlocale(locale.LC_NUMERIC, "en_US.UTF-8")
```
bouweandela: Or should I use the `"C"` locale? I'm not sure if `"en_US"` is installed on every system.

bouweandela: Changed in b1a9302

valeriupredoi:
hi folks, this would be nice if you guys had a mo to approve and get it in the new 3.11 :beer:

valeriupredoi (approved):

pp-mo:
I'm guessing the fails will be fixed by rebasing onto current main.
So I'll merge that in and we'll see how it goes...

bouweandela:
All checks are green now :rocket:

pp-mo:
> All checks are green now 🚀

Great work @bouweandela  and sorry it has taken so long to respond !

One minor follow-on question : Is there any reasonable way of testing this to show it really works ?
( If not, I don't think we need to insist on it. )

bouweandela:
>  Is there any reasonable way of testing this to show it really works ?

Testing that it really works would require installing a locale with the issue on the test system, e.g. `nl_NL.utf8` and then running
```
export LC_NUMERIC='nl_NL.utf8'
python -c 'import locale; locale.setlocale(locale.LC_ALL, ""); import cf_units'
```
so that seems unpractical.

As an alternative, I tried adding a test in `cf_units/tests/unit/test__udunits2.py` that the locale is not set when reading the udunits database, but it's a bit tricky because the code is executed on import. When run separately this test seems to work, but then many other tests fail because I am unable to restore the original state after the test. Here is what I got so far:
```python
def test_import(monkeypatch):
    """Test that locale is not set.

    The locale should not be set when creating a system object at module
    import time. See https://github.com/SciTools/cf-units/issues/435 for
    additional information.
    """

    def mock_read_xml(*args, **kwargs):
        assert locale.getlocale(locale.LC_NUMERIC) == (None, None)

    with monkeypatch.context():
        monkeypatch.setattr(_ud, "read_xml", mock_read_xml)
        locale.setlocale(locale.LC_ALL, "")
        importlib.reload(cf_units)

    importlib.reload(cf_units)
```
when running
```
pytest cf_units/tests/unit/test__udunits2.py::test_import cf_units/tests/unit/unit/test_Unit.py::Test_change_calendar::test_no_change
```
the second test fails, while separately they pass fine. Would you know how to restore the state after `test_import`?

pp-mo:
OK this sounds like too much trouble to test -- the relevant scope is not appropriate for a unit test anyway.
So, I have confirmed locally that this works + I'm just going to merge.

pp-mo (approved):

PR #458, opened 2024-09-18
test macos wheels

HGWright:


.github/workflows/ci-wheels.yml
```diff
@@ -20,46 +20,55 @@ on:
...
+  workflow_dispatch:
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   build_bdist:
     name: "Build ${{ matrix.os }} (${{ matrix.arch }}) wheels"
+    # TBD: extend to support windows
```
ESadek-MO, line 32: Should this be a To Be Decided, or a To Do?

ESadek-MO (approved):
I'm happy with this, just one question then happy to go!

.github/workflows/ci-wheels.yml
HGWright: It seems that windows has never been enabled here, so its not a clear TODO: so I think TBD is appropriate.

PR #460, opened 2024-09-19
Adopt cython3 (require >=3)

### LINKED ISSUE #360: Support Cython v3
( was formerly : "**drop cython language_level on v3.0.0**" )

## 📰 Custom Issue

Once `cython` 3.0.0 lands:

- drop use of the `cython` `language_level` compiler directive (`3str` will be the default behaviour)
  - #358
- add a min pin for `cython>=3.0` for `conda` and PyPI deps
- address `warning: #warning "Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]`, see
  - https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html?highlight=NPY_NO_DEPRECATED_API#numpy-c-api 
  - https://cython.readthedocs.io/en/latest/src/tutorial/numpy.html?highlight=import_array#adding-types

See https://github.com/cython/cython/milestone/58

pp-mo:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
   "setuptools_scm[toml]>=7.0",
   "wheel",
   "oldest-supported-numpy",
-  "Cython"
+  "Cython>=3.0"
 ]
 # Defined by PEP 517
 build-backend = "setuptools.build_meta"
@@ -73,3 +73,6 @@ line_length = 79
 profile = "black"
 skip_gitignore = "True"
 verbose = "False"
+
+[tool.cibuildwheel]
+before-all = "uname -a"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,3 @@ line_length = 79
 profile = "black"
 skip_gitignore = "True"
 verbose = "False"
-
-[tool.cibuildwheel]
-before-all = "uname -a"
--- a/setup.py
+++ b/setup.py
@@ -144,6 +144,8 @@ if cythonize:
     [udunits_ext] = cythonize(
         udunits_ext,
         compiler_directives=COMPILER_DIRECTIVES,
+        # Assert python 3 source syntax. Default for Cython>=3, but let's be definite.
+        language_level="3str",
     )
 
 cmdclass = {"clean_cython": CleanCython, "build_ext": numpy_build_ext}
--- a/setup.py
+++ b/setup.py
@@ -144,7 +144,8 @@ if cythonize:
     [udunits_ext] = cythonize(
         udunits_ext,
         compiler_directives=COMPILER_DIRECTIVES,
-        # Assert python 3 source syntax. Default for Cython>=3, but let's be definite.
+        # Assert python 3 source syntax: Currently required to suppress a warning,
+        #  even though this is now the default (as-of Cython v3).
         language_level="3str",
     )
 
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,6 @@ if cythonize:
     [udunits_ext] = cythonize(
         udunits_ext,
         compiler_directives=COMPILER_DIRECTIVES,
-        language_level="3str",
     )
 
 cmdclass = {"clean_cython": CleanCython, "build_ext": numpy_build_ext}
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,9 @@ except ImportError:
     cythonize = False
 
 COMPILER_DIRECTIVES = {}
-DEFINE_MACROS = None
+# This Cython macro disables a build warning, obsolete with Cython>=3
+#  see : https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api
+DEFINE_MACROS = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
 FLAG_COVERAGE = "--cython-coverage"  # custom flag enabling Cython line tracing
 BASEDIR = Path(__file__).resolve().parent
 PACKAGE = "cf_units"
@@ -119,7 +121,7 @@ def numpy_build_ext(pars):
 
 if FLAG_COVERAGE in sys.argv or environ.get("CYTHON_COVERAGE", None):
     COMPILER_DIRECTIVES = {"linetrace": True}
-    DEFINE_MACROS = [("CYTHON_TRACE", "1"), ("CYTHON_TRACE_NOGIL", "1")]
+    DEFINE_MACROS += [("CYTHON_TRACE", "1"), ("CYTHON_TRACE_NOGIL", "1")]
     if FLAG_COVERAGE in sys.argv:
         sys.argv.remove(FLAG_COVERAGE)
     print('enable: "linetrace" Cython compiler directive')
--- a/requirements/cf-units.yml
+++ b/requirements/cf-units.yml
@@ -7,7 +7,7 @@ dependencies:
 # setup dependencies
   - setuptools >=45
   - setuptools_scm >=7.0
-  - cython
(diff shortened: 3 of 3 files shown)


pp-mo:
@trexfeathers  I think this may possibly be all that is needed ?

So far, this adopts exactly what Bill proposed : 
  - insist on Cython>=3
  - remove the "language_level" setting.

But .. I'm still not sure it wouldn't be better to simply _unpin_ Cython, and leave the language-level explicit (which is then not superfluous).

Particularly as I see that the build does raises a warning about the default having changed recently.  
`/Users/runner/work/cf-units/cf-units/.tox/py311-osx-test/lib/python3.11/site-packages/Cython/Compiler/Main.py:381: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /Users/runner/work/cf-units/cf-units/cf_units/_udunits2.pxd`

What do you think ?

pp-mo:
Updated to enable macos-14 (hopefully 🤞 )

pp-mo:
> Updated to enable macos-14 (hopefully 🤞 )

Initial attempt got a failure from setuptools-scm, which failed to identify the version.
On closer inspection of the log, it seems to be a github access timeout.

But it fails again on retry.  Can't easily work out whether it _is_ essentially a timeout, or maybe a timeout due to something else failing.

pp-mo:
Updated again to pick up the git-path fix (i.e. #464)
Hopefully that gives a solid pass (where before it seemed variable with lots of re-spins)

pyproject.toml
```diff
@@ -5,7 +5,7 @@ requires = [
   "setuptools_scm[toml]>=7.0",
   "wheel",
   "oldest-supported-numpy",
-  "Cython"
+  "Cython>=3.0"
```
trexfeathers, line 8: > @trexfeathers I think this may possibly be all that is needed ?
> 
> So far, this adopts exactly what Bill proposed :
> 
> * insist on Cython>=3
> * remove the "language_level" setting.
> 
> But .. I'm still not sure it wouldn't be better to simply _unpin_ Cython, and leave the language-level explicit (which is then not superfluous).
> 
> Particularly as I see that the build does raises a warning about the default having changed recently. `/Users/runner/work/cf-units/cf-units/.tox/py311-osx-test/lib/python3.11/site-packages/Cython/Compiler/Main.py:381: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /Users/runner/work/cf-units/cf-units/cf_units/_udunits2.pxd`
> 
> What do you think ?

We're talking a lot about maintenance burden at the moment. Bugs in cf-units that are specific to `cython<3` could be a real maintenance headache. TLDR just because software _can_ support older versions doesn't mean that human developers _want_ to:

- They are increasingly likely as we continue to develop cf-units using the latest releases of Cython, while user transition may be more gradual.
- It would be difficult for us to replicate and diagnose a reported bug like this - unless we have gathered a particularly detailed bug report we will probably be doing our testing using the latest version of Cython.
- Once we were aware that the bug was not present with `cython>=3`, we would be reluctant to fix it anyway. Including the minimum pin saves everyone time, and possibly saves disagreements about 'false advertising'.

**I would like to include this minimum pin.**

setup.py
```diff
@@ -142,7 +142,6 @@ def _set_builtin(name, value):
     [udunits_ext] = cythonize(
         udunits_ext,
         compiler_directives=COMPILER_DIRECTIVES,
-        language_level="3str",
```
trexfeathers, line 145: > @trexfeathers I think this may possibly be all that is needed ?
> 
> So far, this adopts exactly what Bill proposed :
> 
> * insist on Cython>=3
> * remove the "language_level" setting.
> 
> But .. I'm still not sure it wouldn't be better to simply _unpin_ Cython, and leave the language-level explicit (which is then not superfluous).
> 
> Particularly as I see that the build does raises a warning about the default having changed recently. `/Users/runner/work/cf-units/cf-units/.tox/py311-osx-test/lib/python3.11/site-packages/Cython/Compiler/Main.py:381: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /Users/runner/work/cf-units/cf-units/cf_units/_udunits2.pxd`
> 
> What do you think ?

Since one of the other suggestions in #360 was specifically focussed on suppressing a warning (the deprecated NumPy one), I'd say that it's equally worth including `language_level` here, **and a comment explaining** - it might seem redundant to someone coming in cold later.

pyproject.toml
```diff
@@ -73,3 +73,6 @@ line_length = 79
 profile = "black"
 skip_gitignore = "True"
 verbose = "False"
+
+[tool.cibuildwheel]
+before-all = "uname -a"
```
trexfeathers: What's up with this? Is it a hangover from the CI problems?

pyproject.toml
trexfeathers, line 1: I don't have a good place to put this request 😅

Please could you work out where to put the [Numpy warning suppression](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api).

The warning is currently being raised during our CI, so we'll know when we get the suppression right. I have done the due diligence to confirm that it is right for us to suppress the warning - it's precautionary and is NOT evidence that our code is calling something it shouldn't.

trexfeathers (changes requested):

pyproject.toml
pp-mo: OK then that just means we leave it as it is.

pyproject.toml
pp-mo: No it's been there a while.
TBH I took a quick look + concluded I didn't understand it, but perhaps I didn't need to.
I will dig a bit deeper + see if I can work out what it's for...

pp-mo: OK I think this is some junk I mistakenly copied from another implementation.
I will remove it.

setup.py
```diff
@@ -142,6 +144,7 @@ def _set_builtin(name, value):
     [udunits_ext] = cythonize(
         udunits_ext,
         compiler_directives=COMPILER_DIRECTIVES,
+        # Assert python 3 source syntax. Default for Cython>=3, but let's be definite.
```
trexfeathers: My main motive is to suppress a warning, rather than 'belt and braces'. Would you be OK mentioning that?

setup.py
pp-mo: See https://github.com/SciTools/cf-units/pull/460/commits/a53ae2012cf3c98e644141bad5dcd05235d481c7

pyproject.toml
pp-mo: See [a7f476b](https://github.com/SciTools/cf-units/pull/460/commits/a7f476b32f7f1308f63a4022ab6d43a8e3c5096a)

A search for "npy" in the "CI wheels/build" logs shows that this warning **_was_** being raised, but not any more.

pyproject.toml
pp-mo: See https://github.com/SciTools/cf-units/pull/460/commits/55ba22182a689ad76c7f7dde48a2fff278664ae8

pyproject.toml
trexfeathers: Can confirm this has worked 👍

trexfeathers (changes requested):

trexfeathers (approved):

setup.py
pp-mo: Sure, that's still a sufficient reason. https://github.com/SciTools/cf-units/pull/460/commits/f2012517e4a35376b8c2b364652ea5a417319dbd

PR #467, opened 2024-09-20
[DOCS] Updated docstring for date2num

### LINKED ISSUE #202: Wrong dtype of return in date2num

ukmo-ccbunney:
## 🚀 Pull Request
The underlying `cftime` package will now preferentially return an integer rather than a float if if encoded datetime(s) can be expressed exactly as an integer.

### Description
Docstring for `num2date` updated to reflect this underlying change in behaviour.
Extra example added showing outputs of type integer where encoded allows.

Fixes: #202

### DIFF
--- a/cf_units/__init__.py
+++ b/cf_units/__init__.py
@@ -1861,6 +1861,10 @@ class Unit(_OrderedHashable):
         Works for scalars, sequences and numpy arrays. Returns a scalar
         if input is a scalar, else returns a numpy array.
 
+        Return type will be of type `integer` if (all) the times can be
+        encoded exactly as an integer with the specified units,
+        otherwise a float type will be returned.
+
         Args:
 
         * date (datetime):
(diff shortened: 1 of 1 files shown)


pp-mo (approved):
:+1: aced it !
🎯

PR #455, opened 2024-09-16
Fixes for CI wheels

HGWright:

PR #470, opened 2024-09-20
Fixlocks

pp-mo:
Addresses #466

From the actions panel, it appears that the lockfile updates were successfully occurring with the previous py39/py310/py311 version.
But this broke for python 3.12 versions, when python versions were updated.

**Also** ... Prior to the Python version update, the lockfiles update action seemed to be working, but did not generate PRs as expected.
It may be simply that the target branch "" had been removed, and the actions have permission to update that, but not to (re-)create it.  We will check that subsequently ...

PR #472, opened 2024-09-20
Pin numpy to v1, but no tighter.

pp-mo:
I think pinning numpy to exactly 1.26.4 was probably an oversight, introduced with #455


requirements/cf-units.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
 # core dependencies
   - antlr-python-runtime 4.11.1.*  # To update this, see cf_units/_udunits2_parser/README.md
   - cftime>=1.2
-  - numpy==1.26.4
+  - numpy<1.99
```
trexfeathers: I think this is the traditional syntax!

```suggestion
  - numpy<2
```

trexfeathers:
Superceded by #470

PR #481, opened 2024-09-24
Fix ci-tests env generation, align with ci-locks.

pp-mo:
Adopting the changes which appeared to be necessary for the ci-locks action in #470 :  
Do the same to ci-tests.


pp-mo:
Ping @trexfeathers @bjlittle  can we accept this now ?

Big 🥕 it also stops using conda-main !

bjlittle (approved):

pp-mo:
Thanks @bjlittle !

PR #482, opened 2024-09-24
Break doctest.

pp-mo:
Just checking that the doctest aspect of #467 functions as expected (i.e. as a test of the revised docstring + behaviour)


pp-mo:
Confirmed, the doctest functions as a test,  so that's all we need.

But, see https://github.com/SciTools/cf-units/issues/202#issuecomment-2372251145
which means there may still be something to do here

PR #483, opened 2024-09-25
[DOCS] Update docstring for date2num (module function)

### LINKED ISSUE #202: Wrong dtype of return in date2num

ukmo-ccbunney:
## 🚀 Pull Request
As per #467, but for module function `date2num`

The underlying cftime package will now preferentially return an integer rather than a float if if encoded datetime(s) can be expressed exactly as an integer.

### Description
Docstring for `cf_units.date2num` updated to reflect this underlying change in behaviour.
Extra example added showing outputs of type integer where encoded allows.

#467 fixed docstring for `cf_units.Unit.date2num`, but this module level function was missed.

Fixes: https://github.com/SciTools/cf-units/issues/202


cf_units/__init__.py
```diff
@@ -365,6 +369,13 @@ def date2num(date, unit, calendar):
         ...               cf_units.CALENDAR_STANDARD)
         array([6.5, 7.5])
 
+        # Integer type preferentially returned if possible:
```
pp-mo:
```suggestion
        >>> # Integer type preferentially returned if possible:
```

Now I got an html build to work, I find that this doesn't render correclty
(and neither does the original in #467 !)

It seems not to like the blank line + stops rendering as a code example.
If you prefix with the correctly indented ">>> ", that seems to fix it.

If you don't like that and would prefer a blank line, please do, but I'm not myself sure how to do that.

pp-mo (changes requested):
Can you please fix the Unit.date2num similarly ?

It seems increasingly that I made a mess of reviewing #467 !

cf_units/__init__.py
pp-mo: Oh, on a roll here !!
I think you can use "\<BLANKLINE>" somehow : see [here](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest)

ukmo-ccbunney: Ah yes - I see now that this is doctest expecting the comment line to be an output of the previous test.
Originally, I was running `make doctest` to get Sphinx to run the doctests and this was passing (even though it shouldn't which makes me think it wasn't __really__ running the doctests). 😲

Now I am running the doctests with `pytest --doctest-modules` which correctly catches this error.

I believe that `<BLANKLINE>` is only valid when you **expect** a blank line as a test output, so I think the best solution here is to simply ensure the comment and any blank lines are prepended with  `>>>`. However, for styling consistency, I have removed the blank line (but kept the comment)

I've done this for both the `cftime.date2num` and `cftime.Unit.date2num`

pp-mo (approved):
:+1: thanks @ukmo-ccbunney

trexfeathers:
I have a fix for those pre-commit errors in #484, if you wanted to merge that one first?

pp-mo:
( merge-in fixed the pre-commit failure. -- thanks @trexfeathers for #484 )

PR #427, opened 2024-05-01
Numpy 2 testing

ocefpaf:
### Description
This PR adapts the code for numpy 2.0. I'm building on top of #313 b/c this one should be merged after that one.
Note that I did not add numpy 2.0 to the test matrix yet. The test matrix here is quite complex with automated conda-lock, tox, etc. It is quite hard to contribute to it!

PR #488, opened 2024-09-27
Support numpy 2

stephenworsley:
### Description
Replaces #427


pp-mo (approved):
Great, can't really fault this.  And it seems we didn't need a whole bunch of test fixes.

PR #489, opened 2024-09-27
Read the Docs fixes and modernisation

### LINKED ISSUE #486: Fix docs build + enable on PRs
As for Iris, so we get checks on the docs build via RTD

Should also do a doctest check,

~We already seem to be getting that (cf #482)  "somehow", but I'm not sure how.~
(Update : pytest docs testing is configured [here](https://github.com/SciTools/cf-units/blob/d758886aa55b483b83f92316793ac90903d1ab9c/pyproject.toml#L38-L42) )

trexfeathers:
### Description

The valid format of intersphinx mapping has apparently changed.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
     python: mambaforge-4.10
   jobs:
@@ -14,6 +14,11 @@ build:
       # create a "common" link to the underlying rtd conda environment,
       # which maybe named "latest", "stable" or the cf-units version
       - ln -s ${CONDA_ENVS_PATH}/${CONDA_DEFAULT_ENV} ${CONDA_ENVS_PATH}/common
+      # Need to stash the local changes that Read the Docs makes so that
+      #  setuptools_scm can generate the correct cf-units version.
+      - git stash
+    post_install:
+      - git stash pop
 
 conda:
   # NOTE: the activation script for this environment does not get run.
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,6 +16,10 @@ build:
       - ln -s ${CONDA_ENVS_PATH}/${CONDA_DEFAULT_ENV} ${CONDA_ENVS_PATH}/common
 
 conda:
+  # NOTE: the activation script for this environment does not get run.
+  #  Therefore, the UDUNITS2 environment variables are not set.
+  #  They are instead set via the RTD dashboard (not yet possible via this
+  #  configuration file - readthedocs/readthedocs.org#6311).
   environment: requirements/cf-units.yml
 
 sphinx:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -28,5 +28,4 @@ release = current_version
 
 html_theme = "alabaster"
 
-# Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {"http://docs.python.org/": None}
+intersphinx_mapping = {"python": ("http://docs.python.org/3/", None)}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -27,5 +27,8 @@ release = current_version
 # -- Options for HTML output ----------------------------------------------
 
 html_theme = "alabaster"
+html_theme_options = {
+    "description": f"version {release}",
+}
 
 intersphinx_mapping = {"python": ("http://docs.python.org/3/", None)}


.readthedocs.yml
```diff
@@ -14,8 +14,17 @@ build:
       # create a "common" link to the underlying rtd conda environment,
       # which maybe named "latest", "stable" or the cf-units version
       - ln -s ${CONDA_ENVS_PATH}/${CONDA_DEFAULT_ENV} ${CONDA_ENVS_PATH}/common
+      # Need to stash the local changes that Read the Docs makes so that
+      #  setuptools_scm can generate the correct cf-units version.
+      - git stash
+    post_install:
+      - git stash pop
```
trexfeathers, line 21: See examples in other repos:

- SciTools/iris#4926
- SciTools/iris-grib#444

docs/source/conf.py
```diff
@@ -28,5 +28,4 @@
 
 html_theme = "alabaster"
 
-# Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {"http://docs.python.org/": None}
+intersphinx_mapping = {"python": ("http://docs.python.org/3/", None)}
```
trexfeathers, line 34: Taken from Iris:

https://github.com/SciTools/iris/blob/38a8ab9f0ecedc481e74d74553bf85b93e8b0603/docs/src/conf.py#L257

pp-mo (dismissed):
Happy with this + I have reviewed the docs on the PR+ see it is working OK

However I note that, despite your including a fix to avoid the post-tagged version number, the docs front page itself doesn't show the version. 
Can we fix that easily ?

.readthedocs.yml
```diff
@@ -14,8 +14,17 @@ build:
...
 
 conda:
+  # NOTE: the activation script for this environment does not get run.
+  #  Therefore, the UDUNITS2 environment variables are not set.
+  #  They are instead set via the RTD dashboard (not yet possible via this
+  #  configuration file - readthedocs/readthedocs.org#6311).
```
trexfeathers, line 27: I have not changed anything here, but I found it difficult to find out what was going on so a comment seemed helpful for future developers.

trexfeathers:
>However I note that, despite your including a fix to avoid the post-tagged version number, the docs front page itself doesn't show the version.
>Can we fix that easily ?

@pp-mo I have sorted this - [see this example](https://cf-units--489.org.readthedocs.build/en/489/):

image

Note that the version is of course meaningless when building from a pull request, but this at least demonstrates where the information will be displayed. The lack of `dirty` in the version name also demonstrates that the `git stash` approach is working.

pp-mo:
> @pp-mo I have sorted this - [see this example](https://cf-units--489.org.readthedocs.build/en/489/):
> Note that the version is of course meaningless when building from a pull request, but this at least demonstrates where the information will be displayed. The lack of `dirty` in the version name also demonstrates that the `git stash` approach is working.

I don't think this is correct, since it reports "0.1dev<xxx>", but it should be "3.2" or possibly "3.3".
I think [this code](https://github.com/SciTools/cf-units/blob/3acf02b61293867d35b836f8898f094d24beffda/docs/source/conf.py#L23-L25) is the place it comes from, not clear why that would be wrong...

trexfeathers:
> > @pp-mo I have sorted this - [see this example](https://cf-units--489.org.readthedocs.build/en/489/):
> > Note that the version is of course meaningless when building from a pull request, but this at least demonstrates where the information will be displayed. The lack of `dirty` in the version name also demonstrates that the `git stash` approach is working.
> 
> I don't think this is correct, since it reports "0.1dev", but it should be "3.2" or possibly "3.3". I think [this code](https://github.com/SciTools/cf-units/blob/3acf02b61293867d35b836f8898f094d24beffda/docs/source/conf.py#L23-L25) is the place it comes from, not clear why that would be wrong...

It is definitely correct. As I say - pull request builds are meaningless in this regard.

trexfeathers:
If it helps you feel more confident, here is what I get if I build locally:

image

EDITED: update image for my very latest install

docs/source/conf.py
ukmo-ccbunney: Os the intersphinx_mapping actually generating an cross links to numpy? I couldn't find any in the generated docs.
I assume this is just a fix to stop it from breaking the docs build?

trexfeathers: @ukmo-ccbunney this mapping is for Python, rather than NumPy. You're right that no cross-links are currently being rendered in the docs. The only usage is `approx_equal`, which is both un-documented and deprecated:

https://github.com/SciTools/cf-units/blob/65e545c971f96eeddde9fdb45386838756410288/cf_units/util.py#L15-L21

... but it does show that there is precedent for using a Python intersphinx so I would be reluctant to remove it in case it's useful to future developers.

Removing intersphinx entirely would not break the docs build, if you felt strongly that we should do that.

ukmo-ccbunney: Apologies - I meant Python (I saw numpy in the Iris example and typed that instead)
Happy to leave it - perhaps we'll add some more intersphinx links in the future (like for numpy!)

ukmo-ccbunney (approved):

PR #491, opened 2024-09-27
Use unshallow Git fetch for setuptools-scm (Read the Docs)

trexfeathers:
### Description

@pp-mo was right - there is an issue with the information setuptools_scm is getting when running on Read the Docs. Clearly a gap in my knowledge there, sorry!

Adding the same `git fetch --unshallow` that Iris has seems to fix the issue - can see that the RTD build on this PR has a correct version, whereas the current `main` does not.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,6 +6,9 @@ build:
     python: mambaforge-4.10
   jobs:
     post_checkout:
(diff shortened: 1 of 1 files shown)


HGWright (approved):

HGWright:
Thanks @trexfeathers

PR #497, opened 2024-10-02
Use the Codecov GitHub Action

trexfeathers:
### Description

The current upload routine - using the CLI - doesn't seem to be working. Several of our other repos use the GitHub Action so I thought I would try that.


pp-mo (approved):
All seems good :smiley:

PR #461, opened 2024-09-19
Fully support Python 3.12

trexfeathers:
### Description

See https://github.com/SciTools/cf-units/pull/455/files#r1765037237 - #455 was missing a few changes.

I have also introduced a test that will catch things like this in future.

The commit history includes some narrative for the changes I have made.

PR #512, opened 2024-10-30
Adopt Python v3.13

### LINKED ISSUE #506: Support Python 3.13
3.13 was [released 2024-10-07](https://peps.python.org/pep-0719/#lifespan)
Aligning with [SPEC0](https://scientific-python.org/specs/spec-0000/), this would mean we support 3.11, 3.12, 3.13 + drop 3.10.
Presumably, that is ok.

pp-mo:
Closes #506

Modelled on #461, frankly.

It's too soon to drop Python 3.10.
Also may not work, given problems encountered in https://github.com/SciTools/iris/pull/6195

But it's certainly worth a go.


pp-mo:
Frankly I didn't expect that to work !

But now it seems all good
 \-- unless I misunderstood something like `test_python_versions` ?? 
Could you maybe cast an eye over this, @trexfeathers ?

trexfeathers (approved):

pp-mo:
Thanks @trexfeathers !

Issue #521, opened 2024-11-14
Address failing numpydoc tests


ukmo-ccbunney:
The following numpydoc errors are being generated and need addressing:

```[tasklist]
### Numpydoc error codes
- [ ] `GL08`  The object does not have a docstring
- [ ] `PR01` Parameters {xxxx} not documented
- [ ] `RT01` No Returns section found
- [ ] `SS03` Summary does not end with a period
- [ ] `SS06` Summary should fit in a single line
- [ ] `SS05` Summary must start with infinitive verb, not third person (e.g. use "Generate" instead of "Generates")
```

These are currently being set to be ignored in the `pyproject.toml`

Issue #520, opened 2024-11-14
Address failing mypy tests


ukmo-ccbunney:
The following typing errors are being generated by `mypy` and need addressing in the code:

```[tasklist]
### Error code
- [ ] `arg-type`
- [ ]  `assignment`
- [ ]  `attr-defined`
- [ ]  `func-returns-value`
- [ ]  `misc`
- [ ]  `no-untyped-call`
- [ ]  `no-untyped-def`
- [ ]  `operator`
- [ ]  `redundant-expr`  Note: this one should be explicitly enabled in `enable_error_code`
- [ ]  `unreachable`
- [ ]  `var-annotated`
```

Currently, failing type checks are added to the `disable_error_code` lust in `pyproject.toml`

Note, the `cf_units/_udunits2_parser` is being ignored and is out-of-scope for this Issue.

Issue #462, opened 2024-09-19
Address failing tests in repo-review


ukmo-ccbunney:
PR #456 added `repo-review` as a pre-commit hook using the `sp-repo-review` plugin (https://pypi.org/project/sp-repo-review/)

Several of the checks are failing for the `cf-units` repository (expected). The failing tests are currently set in an ignore list in `pyproject.toml` to avoid failures in the workflow, but it would be good practice if we could address them:

```[tasklist]
### Test
- [ ] PY005: Has tests folder
- [ ] PP003: Does not list wheel as a build-dep
- [x] PP304: Sets the log level in pytest
- [x] PP305: Specifies xfail_strict
- [x] PP306: Specifies strict config
- [x] PP307: Specifies strict markers
- [ ] PP309: Filter warnings specified
- [x] GH104:  Use unique names for upload-artifact
- [ ] GH212: Require GHA update grouping
- [x] PC110: Uses black or ruff-format
- [x] PC140: Uses a type checker
- [x] PC160: Uses a spell checker
- [ ] PC170:Uses PyGrep hooks (only needed if rST present)
- [ ] PC180:Uses a markdown formatter
- [x] PC190: Uses Ruff
- [ ] PC901: Custom pre-commit CI message
- [x] MY100:  Uses MyPy (pyproject config)
- [ ] MY103: Warn unreachable
- [ ] MY105: MyPy enables redundant-expr
- [x] RF001: Uses RUFF
```

We may decide some checks are not relevant, in which case they can be removed from this above TODO list and left in the `ignore` list in `pyproject.toml`

ukmo-ccbunney:
Several tests resolved in #522

PR #522, opened 2024-11-15
Initial setup of QA pre-commit tools [cf-units]

ukmo-ccbunney:
### Description
Added following pre-commit hooks:
  - Extra "default" hooks from pre-commit (trailing whitespace, etc)
  - MyPy
  - codespell
  - validate_pyproject
  - numpydoc

`sp-repo-review` was already setup for cf-units, but have updated the ignore list as various tests now pass.

`ruff` was enabled, but with only selected tests. Have enabled "all"  tests.

For all QA tools above, the results of any "auto-fixes" have been included in this repo, mainly fixes for:
  - trailing whitespace
  - missing newline at end of file
  - docstring reformatting
  - some code style linting from ruff

All other failing tests have been added to ignore lists. See separate issues:
   - #520 
   - #521
   - #526 
   - #462


cf_units/_udunits2_parser/_antlr4_runtime/Recognizer.py
```diff
@@ -7,7 +7,7 @@
 from .RuleContext import RuleContext
 from .Token import Token
 
-# need forward delcaration
+# need forward declaration
```
pp-mo: It's not appropriate to modify **anything** inside `cf_units._udunits_parser._antlr4_runtime` ,
since this code is all copied directly from another project.

Effectively this is generated code, derived from the action of [compile.py](https://github.com/SciTools/cf-units/blob/v3.3.0/cf_units/_udunits2_parser/compile.py).
In a more ideal world, this code wouldn't exist in this repo, and we'd have it as a importable dependency instead.
It's only included in the project repo for convenience, specifically to avoid an awkward dependency.
See [./_udunits2_parser/README.md#generating-the-parser-from-the-grammar](https://github.com/SciTools/cf-units/blob/v3.3.0/cf_units/_udunits2_parser/README.md#generating-the-parser-from-the-grammar) for the lowdown.

So, we need to exclude this from the relevant checks.
I _think_ you can just add a top-level [exclude](https://pre-commit.com/#top_level-exclude) control to `.pre-commit-config.yaml`.

.pre-commit-config.yaml
```diff
@@ -2,6 +2,10 @@
 # See https://pre-commit.com/hooks.html for more hooks
 # See https://github.com/scientific-python/cookie#sp-repo-review for repo-review
 
+ci:
```
pp-mo, line 7: For consistency with the "repo procedures" templates in the project, 
I think this should also have a reference line `# See https://pre-commit.ci/#configuration` in the initial comment lines section above.

.pre-commit-config.yaml
```diff
@@ -16,8 +20,21 @@ repos:
     -   id: check-merge-conflict
         # Check for debugger imports and py37+ `breakpoint()` calls in Python source.
     -   id: debug-statements
+        # Check TOML file syntax.
```
pp-mo, line 29: It could be useful to feed these explainers back to the template, with a PR to [the 'template' file in to SciTools/.github](https://github.com/SciTools/.github/blob/6cd87da89ae606ccb4ce0a4c95421c012cb29108/templates/.pre-commit-config.yaml)

ukmo-ccbunney:
@pp-mo I Have resolved the issues I was having with pre-commit.ci.
I managed to reproduce the pre-commit.ci behaviour on my personal laptop which gave me a chance to investigate the issue.

I have reinstated the `PY005` check ("has test folder") to the to `[tool.repo-review] ignore` list.

The MyPy behaviour was more difficult to debug. It turns out that adding files to the MyPy `exclude` list only stops them being passed as filenames to the MyPy call. However, these will **still** be checked if they are `import`ed from some other module that is being checked.

There is a flag that can be passed to stop this `--follow-imports=silent`, but this is not supported via `pyproject.toml`.
However, I've found it can be set using the following too MyPy section:
```
[[tool.mypy.overrides]]
module = [
        "cf_units/_udunits2_parser/parser.*",
        "cf_units/_udunits2_parser/_antlr4_runtime.*",
]
ignore_errors = true
```

The combination of the above fix the errors on my personal laptop and the pre-commit.ci service, so this is ready for review now.

I don't know why the behaviour is different on the MO VDIs though (pre-commit and hook version numbers are same!)

pp-mo:
> `--follow-imports=silent`, but this is not supported via `pyproject.toml`

I don't think turning this totally off is a good idea, anyway.
Your approach to exclude certain files seems better to me.

pyproject.toml
```diff
@@ -69,6 +69,10 @@ Discussions = "https://github.com/SciTools/cf-units/discussions"
 Issues = "https://github.com/SciTools/cf-units/issues"
 Documentation = "https://cf-units.readthedocs.io"
 
+[tool.codespell]
+#ignore-words-list = "Wirth"
```
pp-mo: What is this bit, can we either remove or explain ?

pyproject.toml
```diff
@@ -109,33 +119,121 @@ local_scheme = "dirty-tag"
...
 ]
-line-length = 79
+line-length = 88
+
+[tool.ruff.format]
+preview = false
 
 [tool.ruff.lint]
+ignore = [
+    # NOTE: Non-permanent exclusions should be added to the ".ruff.toml" file.
```
pp-mo: Sorry to hit you with this, but I think the sprint team has now definitely decided **_not_** to do this. 
So, should remove this : the approach at [line 145](https://github.com/SciTools/cf-units/blob/7e9a9bb34bbcc3219391682cf12e7cebaf16dadd/pyproject.toml#L145) is now the way !

pyproject.toml
```diff
@@ -145,22 +243,110 @@ convention = "numpy"
...
-  "PC170",  # Uses PyGrep hooks (only needed if rST present)
-  "PC180",  # Uses a markdown formatter
-  "PC190",  # Uses Ruff
-  "PC901",  # Custom pre-commit CI message
-  "MY100",  # Uses MyPy (pyproject config)
-  "RF001",  # Uses RUFF
+    # TODO: exceptions that still need investigating are below.
+    # Might be fixable, or might become permanent (above):
+    "PY005",  # Has tests folder
+    "PP003",  # Does not list wheel as a build-dep
```
pp-mo: FWIW I think it might be useful to keep these in alphabetical order.
Look like that is true of the existing, anyway

pyproject.toml
```diff
@@ -109,33 +119,121 @@ local_scheme = "dirty-tag"
...
-    # flake8-comprehensions
-    "C4",
-    # isort
-    "I",
-    # pyupgrade
-    "UP",
-]
-ignore = ["B904", "F403", "F405"]
+    "ALL",
+    "D212",  # Multi-line docstring summary should start at the first line
```
pp-mo, line 213: The D212 entry appears to be redundant ?

pyproject.toml
```diff
@@ -109,33 +119,121 @@ local_scheme = "dirty-tag"
...
-    "B",
-    # flake8-comprehensions
-    "C4",
-    # isort
-    "I",
-    # pyupgrade
-    "UP",
-]
-ignore = ["B904", "F403", "F405"]
+    "ALL",
```
pp-mo, line 210: It seems odd to me to have "exclude" followed by "select", since 'exclude' effectively qulaifies the 'select' choices.  
( but I've now raised this offline in our Teams sprint chat )

pp-mo (changes requested):
Ok but a few nit-picky suggestions.

You should also create some PRs to track unfinished business.
Suggest one each for :
  * re-order pre-commit entries based on [updated tempate file](https://github.com/SciTools/.github/pull/29)  : (can do this one right now)
  * mypy
  * ruff
  * numpydocs-validation

cf_units/_udunits2_parser/_antlr4_runtime/Recognizer.py
ukmo-ccbunney: Rolled back changes to files in `_antlr4_runtime` directory and added top level exclusion in `pre-commit-config.yaml`; see 89afe5a

.pre-commit-config.yaml
ukmo-ccbunney: Good spot; fixed in 2749787

pyproject.toml
ukmo-ccbunney: Redundant exception - now removed; 843e6c2

pyproject.toml
ukmo-ccbunney: Yep! Old comment. I've got nothing in .ruff.toml anyhow. Removed: 5a955ff

pyproject.toml
ukmo-ccbunney: This was explicitly set in the standard scitools linter config (https://github.com/orgs/SciTools/projects/43/views/3?pane=issue&itemId=85886972), so I've left it in?

pyproject.toml
ukmo-ccbunney: Will raise a separate PR for this once we have a decision?

pyproject.toml
ukmo-ccbunney: They were in order that repo-review reports them but it makes sense to alphabetise them for consistency; see 01fe1b7

.pre-commit-config.yaml
ukmo-ccbunney: I don't follow you here - the comments already exist in the template file.

ukmo-ccbunney:
> You should also create some PRs to track unfinished business. Suggest one each for :

I've already created Issues for these (not PRs), see links in description

> * re-order pre-commit entries based on [updated tempate file]
(https://github.com/SciTools/.github/pull/29)  : (can do this one right now)

Done; see d831bfb

.pre-commit-config.yaml
pp-mo: Oh I hadn't seen that !

pyproject.toml
pp-mo: Yes, but I'm pretty sure it's wrong all the same.

To my frustration, they don't seem to actually document the "ALL" magic 😞 
But IIUC, the list of 'select' options is additive, so including "ALL" means anything else is redundant by definition.

(NB we have an ongoing offline teams convo about this)

pp-mo: > they don't seem to actually document the "ALL" magic

Oh my mistake, it is there in the [command-line help](https://docs.astral.sh/ruff/configuration/#full-command-line-interface)

pp-mo (changes requested):
I think removing "D212" from the ruff.lint `select` is worth doing,
not least because the comment is confusing + I think wrong!

otherwise all looks great now + good to go !

pyproject.toml
ukmo-ccbunney: D212 removed from Ruff `select`; bb551e1
This resulted in a "unused NOQA" autofix on `__init__.py` which is also included in the commit.

ukmo-ccbunney:
Right - `D212` is back in the Ruff `select` list after `ALL`.
Hopefully this is everything that is needed now for this PR.

pp-mo (approved):

PR #530, opened 2024-11-22
[cf-units] Sorted pyproject.toml

ukmo-ccbunney:
### Description
Used `sort-toml` utility with some minor manual changes afterwards to preserve grouping of elements in arrays.

This should results in all sections and keys being sorted alphabetically.

Order of array elements has not been changed.


pp-mo (changes requested):
I guess this is all OK, 
but can we not also adopt the sorting operation within the pre-commit ?

ukmo-ccbunney:
> I guess this is all OK, but can we not also adopt the sorting operation within the pre-commit ?

I didn't adopt anything in the pre-commit as this was one of the outstanding questions from the sprint (what sorting tool we should adopt).

pp-mo (approved):

PR #556, opened 2025-02-13
ci: build windows wheel

trim21:
### Description

build pypi wheel for windows

reason: 

https://github.com/SciTools/iris depend on this, which has a pypi windows wheel.

how:

- install libexpat from vcpkg
- install udunits from source with cmake + vcpkg

old `CIBW_TEST_COMMAND` failed with reason I don't understand, looks like some kind of shell problem... So I only run pytest on windows.

---

Also another PR to install udunits from conda, if maintainers prefer https://github.com/SciTools/cf-units/pull/558


ukmo-ccbunney:
Hi @trim21 
Thanks for raising this PR. 🚀
I think your alternative PR #558 using conda is a better solution to this as it avoids pinning the udunits2 version in the `ci-wheels.yml` file.

trim21:
ok

PR #558, opened 2025-02-13
ci: build windows pypi wheels (with conda)

trim21:
### Description

almost same patch with https://github.com/SciTools/cf-units/pull/556 , but install udunits from conda


.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,22 @@ jobs:
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: python -m pytest --pyargs cf_units
```
ukmo-ccbunney: I imagine the original command was failing on Windows (as you mention in #556) due to the `&&` operator not being supported by the Windows shell (at least not in the version provided in the GHA runner).

To keep it consistent with the previous behaviour you could do this:

```suggestion
        CIBW_TEST_COMMAND_WINDOWS: >
          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")
          if ($?) { python -m pytest --pyargs cf_units }
```

or seeing in this case running the tests isn't _really_ dependent on the previous command invocation, just simply running both commands independently is maybe acceptable?

```suggestion
        CIBW_TEST_COMMAND_WINDOWS: >
          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")
          python -m pytest --pyargs cf_units
```

ukmo-ccbunney (changes requested):
Hi @trim21 

Thank you very much for your PR.

I think that this solution using **Conda** to install `udunits2` is probably more robust than building it in the workflow as you proposed in #556. This way, the `udunits2` version is free to update rather than being pinned to a version in the `ci-wheels.yml` file.

I only have one small suggested change with respect to the Windows specific test command, which I will leave to your discretion.

Otherwise I've updated your PR with the latest changes from main and I think we're all good to go on this one. 👍

trim21:
My pr can't trigger testing  unless manually approved by maintainers , I'm not sure if the suggestion work

trim21:
Ok, feel free to cherry pick it to main branch instead of merging PR, if you prefer. That should be able to speedup the review progressing and I'm tired of waiting.

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
...
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")
+          python -m pytest --pyargs cf_units
```
ukmo-ccbunney: My bad - we probably need a `;` here as line breaks are collapsed.
```suggestion
        CIBW_TEST_COMMAND_WINDOWS: >
          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}") ;
          python -m pytest --pyargs cf_units
```

ukmo-ccbunney (changes requested):

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}") ;
```
ukmo-ccbunney: Unterminated string...

```suggestion
          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' ;
```

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' ;
```
ukmo-ccbunney: Apologies - this is still failing.
I think it is down to the quotes used (windows seems to be more picky than linux)
Hopefully this will fix it. Fingers crossed.

```suggestion
          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" ;
```

ukmo-ccbunney (dismissed):
OK. This is all working now.
Apologies for the several iterations to fix the `CIBW_TEST_COMMAND_WINDOWS` value - got there eventually. 🙄
Thanks for the contribution. 🚀

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
...
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" ;
+          python -m pytest --pyargs cf_units
         UDUNITS2_INCDIR: ${{ matrix.incdir }}
         UDUNITS2_LIBDIR: ${{ matrix.libdir }}
         UDUNITS2_XML_PATH: ${{ matrix.xml_path }}
         CIBW_ENVIRONMENT_PASS_LINUX: UDUNITS2_INCDIR UDUNITS2_LIBDIR UDUNITS2_XML_PATH
+        CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: >-
```
ocefpaf, line 110: Is it worth running `delvewheel show {wheel}` too before the repair command?

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
...
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" ;
+          python -m pytest --pyargs cf_units
```
ukmo-ccbunney: Just spotted that this still isn't working - the pytest command is silently being ignored.
I've tested this outside of your PR and I think this should now work:

```suggestion
          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" &&
          python -m pytest --pyargs cf_units
```

I think it was the quotes that were causing the issues originally.

.github/workflows/ci-wheels.yml
trim21: I suggest we don't struggle with multiple command again

.github/workflows/ci-wheels.yml
trim21: then I guess we can merge 2 options of cibuildwheels? CIBW_TEST_COMMAND_WINDOWS and CIBW_TEST_COMMAND

.github/workflows/ci-wheels.yml
```diff
@@ -78,14 +98,24 @@ jobs:
...
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
+        CIBW_BEFORE_BUILD_WINDOWS: python -m pip install delvewheel
         CIBW_TEST_REQUIRES: pytest
         CIBW_TEST_COMMAND: >
           python -c 'import cf_units; print(f"cf-units v{cf_units.__version__}")' &&
           python -m pytest --pyargs cf_units
+        CIBW_TEST_COMMAND_WINDOWS: >
+          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" &&
+          python -m pytest --pyargs cf_units
```
ukmo-ccbunney:
```suggestion
        CIBW_TEST_COMMAND: >
          python -c "import cf_units; print(f'cf-units v{cf_units.__version__}')" &&
          python -m pytest --pyargs cf_units
```

.github/workflows/ci-wheels.yml
ukmo-ccbunney: Agreed. I've made a commit.

ukmo-ccbunney (approved):
OK. Everything is building ok and all tests are running and passing.
Thanks for your patience on this @trim21 - that was much more of a chore sorting out the windows multi-line command than I expected. LGTM.

trim21:
It may be a bad idea to use dll from conda env.

I have a similar PR on other project use c/c++ package from conda env to build a pypi wheel, which cause a segfault on certain case, and we recently reverted it to use a manually build.

PR #577, opened 2025-04-03
Pointless tweak.

pp-mo:


pp-mo:
Just investigating to see what was going on with apparently transient test failures elsewhere

PR #586, opened 2025-04-30
pep639: setuptools license and license-files fields

bjlittle:
### Description

This pull-request mops up some tech debt to comply with [PEP639](https://peps.python.org/pep-0639/) for `setuptools`, also see [license and license-files](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files).

Additionally avoids the following `setuptools` deprecation warning:
```
!!
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:
          License :: OSI Approved :: BSD License
          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
!!
```
See [Deprecated license classifiers](https://peps.python.org/pep-0639/#deprecate-license-classifiers).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,8 +5,8 @@
 build-backend = "setuptools.build_meta"
 # Defined by PEP 518
 requires = [
-    "setuptools>=45",
-    "setuptools_scm[toml]>=7.0",
+    "setuptools>=77.0.3",
+    "setuptools_scm[toml]>=8.0",
     "numpy",
     "Cython>=3.0",
 ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -55,7 +54,8 @@ keywords = [
 name = "cf-units"
 readme = "README.md"
 requires-python = ">=3.10"
-license.file = "LICENSE"
+license = "BSD-3-Clause"
```
pp-mo, line 57: This caused me some concern since the core metadata 'Licence' field is [deprecated](https://peps.python.org/pep-0639/#deprecate-license-field).
But then I eventually found out that the relationship between Core Metadata terms and pyproject.toml keys is non-trivial.
So, in this case, ["licence" --> "Licence Files"](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license)
Likewise "license-files" --> "Licence File", 
and "file" --> "Licence" (deprecated !).

Unexpectedly confusing!

pp-mo (approved):

pp-mo:
Thanks @bjlittle for taking the trouble with this!

PR #605, opened 2025-07-11
Add MANIFEST.in update + ci check

HGWright:
### Description

Been a few years since we updated the MANIFEST.in file. So we are using the new template to do it. We have also since started using this CI check to make sure that the MANIFEST.in file is valid.


trexfeathers:
Replaced by #608

PR #620, opened 2025-09-02
Fix licence header tests -- removed unused complexity.

pp-mo:
This was triggered by trying to fix check failures with #616.
The problem there is that newer "git" has removed the "whatsnew" command.

So, I **_was_** going to adopt https://github.com/SciTools/iris/pull/6672 here
But on closer inspection I discovered that, since we stopped including years in the licence header, the legacy code is completely overcomplicated, since we no longer use the dates at all (N.B. _this is also true in Iris_).
So, this shows a simplification.

Note: it has also been suggested that we **_could_** use "pre-commit ruff" for this, if we adopt [the "CPY001" rule](https://docs.astral.sh/ruff/rules/missing-copyright-notice/).
This idea noted [here in geovista](https://github.com/bjlittle/geovista/pull/532).
**_However_** there's currently an obstacle to that, since "preview" mode is too much work, as discussed [here](https://github.com/astral-sh/ruff/issues/12343)

### DIFF
--- a/cf_units/tests/test_coding_standards.py
+++ b/cf_units/tests/test_coding_standards.py
@@ -3,7 +3,6 @@
 # This file is part of cf-units and is released under the BSD license.
 # See LICENSE in the root of the repository for full licensing details.
 
-from datetime import datetime
 from fnmatch import fnmatch
 from pathlib import Path
 import subprocess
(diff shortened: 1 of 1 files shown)


pp-mo:
## Status
**Put into draft**
Since various inbuilt problems emerged from _**other**_ checks not passing -- notably 
  * wheels build (macos-latest) : suspect the os version here has changed -- see #621 
  * ~docs build also failing (not yet clear why)~ (update: probably a glitch)

HGWright (approved):

PR #627, opened 2025-10-27
Build Wheels with ABI3

### LINKED ISSUE #626: Support Python 3.14
## ✨ Feature Request

3.14 was released in October 2025

## Motivation

That would allow libraries that depend on cf-units also to be installed on Python 3.14. (example: https://github.com/ioos/compliance-checker) 

Is there any plan to release a new version with a new wheel?

ocefpaf:
This PR builds a single wheel for each platform using Python 3.11 ABI3. That means we drop support for 3.10 and 3.9, I can try to re-add 3.10 later but I don't think 3.9 is worth it. I would stick with SPEC 0 though, that way we can remove 3.10 now and make our lives a bit easier.

Note that, with ABI3, we don't need to rebuild the extension every new Python release. I just tested this wheel against 3.14 and it works as expected.

Closes #626 

@scitools-templating: please no share prompt

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -37,16 +37,16 @@ jobs:
           - os: ubuntu-latest
             platform: linux
           - os: ubuntu-latest
-            version: py313
+            version: py314
             posargs: "--cov-report=xml --cov"
           - os: macos-latest
-            version: py313
+            version: py314
             platform: osx
             # On macos, the up-to-date git may not be first on the path
             # N.B. setting includes a final ":", to simplify the path setting command
             gitpath-prepend: "/opt/homebrew/bin:"
           - os: windows-latest
-            version: py313
+            version: py314
             platform: win
       fail-fast: false
     steps:
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -48,6 +48,7 @@ jobs:
           - os: windows-latest
             version: py313
             platform: win
+      fail-fast: false
     steps:
     - name: "Checkout"
       uses: actions/checkout@v6
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        version: [py310, py311, py312, py313]
+        version: [py311, py312, py313, py314]
         gitpath-prepend: [""]
         include:
           - os: ubuntu-latest
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -82,7 +82,10 @@ jobs:
         export PATH=${{ matrix.gitpath-prepend }}$PATH
         which git
         git --version
-        pixi run --frozen pytest ${{ matrix.posargs }}
+        pip install -e . --no-build-isolation --no-deps
+        pushd requirements/
+        pixi run --frozen pytest --pyargs cf_units ${{ matrix.posargs }}
+        popd
 
     - name: "Upload coverage report to Codecov"
       if: contains(matrix.posargs, '--cov')
--- a/.github/workflows/ci-wheels.yml
+++ b/.github/workflows/ci-wheels.yml
@@ -93,7 +93,7 @@ jobs:
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
       uses: pypa/cibuildwheel@v3.3.1
       env:
-        CF_UNITS_LIMITED_API: "1"
+        CIBW_ENVIRONMENT: CF_UNITS_LIMITED_API="1"
         CIBW_SKIP: "pp* *-musllinux*"
         CIBW_ARCHS: ${{ matrix.arch }}
         CIBW_BUILD: "cp311-* cp314-*"
--- a/.github/workflows/ci-wheels.yml
+++ b/.github/workflows/ci-wheels.yml
@@ -93,8 +93,9 @@ jobs:
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
       uses: pypa/cibuildwheel@v3.3.1
       env:
-        CIBW_SKIP: "cp39-* pp* *-musllinux*"
+        CIBW_SKIP: "pp* *-musllinux*"
         CIBW_ARCHS: ${{ matrix.arch }}
+        CIBW_BUILD: "cp311-* cp314-*"
         CIBW_BUILD_FRONTEND: build
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
--- a/.github/workflows/ci-wheels.yml
+++ b/.github/workflows/ci-wheels.yml
@@ -93,6 +93,7 @@ jobs:
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
       uses: pypa/cibuildwheel@v3.3.1
       env:
+        CF_UNITS_LIMITED_API: "1"
         CIBW_SKIP: "pp* *-musllinux*"
         CIBW_ARCHS: ${{ matrix.arch }}
         CIBW_BUILD: "cp311-* cp314-*"
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,12 @@ except ImportError:
 COMPILER_DIRECTIVES = {}
 # This Cython macro disables a build warning, obsolete with Cython>=3
 #  see : https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api
-DEFINE_MACROS = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
+DEFINE_MACROS = [
+    ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
+    # 0x030B0000 -> 3.11
+    ("Py_LIMITED_API", "0x030B0000"),
+    ("CYTHON_LIMITED_API", None),
+]
 FLAG_COVERAGE = "--cython-coverage"  # custom flag enabling Cython line tracing
 BASEDIR = Path(__file__).resolve().parent
 PACKAGE = "cf_units"
@@ -128,8 +133,9 @@ udunits_ext = Extension(
     include_dirs=include_dirs,
     library_dirs=library_dirs,
     libraries=["udunits2"],
-    define_macros=DEFINE_MACROS,
     runtime_library_dirs=(None if sys.platform.startswith("win") else library_dirs),
+    define_macros=DEFINE_MACROS,
+    py_limited_api=True,
 )
 
 if cythonize:
@@ -148,6 +154,7 @@ kwargs = {
     "cmdclass": cmdclass,
     "ext_modules": [udunits_ext],
     "package_data": get_package_data(),
+    "options": {"bdist_wheel": {"py_limited_api": "cp311"}},
 }
 
 setup(**kwargs)
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,11 @@ All other setup configuration is in `pyproject.toml`.
 """
 
 from distutils.sysconfig import get_config_var
-from os import environ
+from os import environ, getenv
 from pathlib import Path
 from shutil import copy
 import sys
+import sysconfig
 
 from setuptools import Command, Extension, setup
 from setuptools.command import build_ext
@@ -19,15 +20,26 @@ try:
 except ImportError:
     cythonize = False
 
+USE_PY_LIMITED_API = (
+    # Builds are not ABI3 default
+    getenv("CF_UNITS_LIMITED_API", "0") == "1"
+    and sys.version_info >= (3, 11)
+    # Free-threaded builds do not support ABI3
+    and not sysconfig.get_config_var("Py_GIL_DISABLED")
+)
+
 COMPILER_DIRECTIVES = {}
 # This Cython macro disables a build warning, obsolete with Cython>=3
 #  see : https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api
-DEFINE_MACROS = [
-    ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
-    # 0x030B0000 -> 3.11
-    ("Py_LIMITED_API", "0x030B0000"),
-    ("CYTHON_LIMITED_API", None),
-]
+DEFINE_MACROS = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
+if USE_PY_LIMITED_API:
+    DEFINE_MACROS.append(
+        [
+            # 0x030B0000 -> 3.11
+            ("Py_LIMITED_API", "0x030B0000"),
+            ("CYTHON_LIMITED_API", None),
+        ]
+    )
 FLAG_COVERAGE = "--cython-coverage"  # custom flag enabling Cython line tracing
 BASEDIR = Path(__file__).resolve().parent
 PACKAGE = "cf_units"
@@ -150,11 +162,16 @@ if cythonize:
 
 cmdclass = {"clean_cython": CleanCython, "build_ext": NumpyBuildExt}
 
+if USE_PY_LIMITED_API:
+    SETUP_OPTIONS = {"bdist_wheel": {"py_limited_api": "cp311"}}
+else:
+    SETUP_OPTIONS = {}
+
 kwargs = {
     "cmdclass": cmdclass,
     "ext_modules": [udunits_ext],
     "package_data": get_package_data(),
-    "options": {"bdist_wheel": {"py_limited_api": "cp311"}},
+    "opention": SETUP_OPTIONS,
 }
 
 setup(**kwargs)
--- a/setup.py
+++ b/setup.py
@@ -166,7 +166,7 @@ kwargs = {
     "cmdclass": cmdclass,
     "ext_modules": [udunits_ext],
     "package_data": get_package_data(),
-    "opention": SETUP_OPTIONS,
+    "options": SETUP_OPTIONS,
 }
 
 setup(**kwargs)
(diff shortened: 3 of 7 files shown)


ocefpaf:
I don't know how tox works to fix the failures here, but a `py310_abl3` wheel is built and tested against py314.

ocefpaf:
I cannot figure out what is wrong with the CIs. In a different env, where the wheels are built and tested, everything passes. I'm guessing it is the lockfiles, maybe the cython version there is different?

.github/workflows/ci-wheels.yml
```diff
@@ -90,9 +78,10 @@ jobs:
       run: conda install udunits2
 
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
-      uses: pypa/cibuildwheel@v3.0.1
+      uses: pypa/cibuildwheel@v3.2.1
       env:
-        CIBW_SKIP: "cp39-* pp* *-musllinux*"
+        CIBW_SKIP: "*-musllinux*"
+        CIBW_BUILD: "cp311-* cp314-*"
```
trexfeathers: This syntax does not specify a range, as far as I can tell ([ref](https://cibuildwheel.pypa.io/en/stable/options/)). Should `312` and `313` be explicitly referenced here?

ocefpaf: Nope. In fact we only need cp311, and that one will be installable in any Python >= 3.11. The cp314 is just for testing purposes so we know nothing will break with future changes in the stable ABI.

trexfeathers: So if I can help this PR to pass its CI, then we'll remove the `314` reference?

ocefpaf: BTW, you can download the wheels generated here in https://github.com/ocefpaf/cf-units/actions/runs/18846733110 and test them in any env with py>=3.11.

ocefpaf: > So if I can help this PR to pass its CI, then we'll remove the `314` reference?

It is nice to have for testing purposes IMO, but it can be removed if you don't care about future proofing.

trexfeathers: As you've probably noticed: there is not enough resource available for maintaining cf-units. So this is really appealing:

>Note that, with ABI3, we don't need to rebuild the extension every new Python release.

Specifying `314` at the moment is good future proofing. But at some point 3.14 will be officially supported[^1] and this setting will be confusing to developers who were not part of this conversation. So yes please lets remove it.

[^1]: apologies if I've misunderstood, I'm having to teach myself cibuildwheel so that I can review this.

ocefpaf: No need to apologize! I'm learning ABI3 as well. Note that, having the 314 there is not a big deal. We do not ship any wheels with it, we only test if the current config works against the latest Python. In theory it should always work b/c the ABI3's promise is stability. However, they may introduce breaking changes every-now-and-then and, by testing against latest Python, we will catch it many years before we are forced to upgrade the min Python build.

trexfeathers: OK I think I'm catching up. You said earlier that we _could_ support 3.10, and 3.11 is just for consistency with SPEC-0, which I'm happy with. So we're talking about this line being a shifting 'window'. Later on it will say:

```
        CIBW_BUILD: "cp312-* cp315-*"
```

If I've got this right, then it would be great to make some appropriate changes to `test_python_versions()`:

https://github.com/SciTools/cf-units/blob/9e34057547f407df84fc3d5bc742f93831cbbe15/cf_units/tests/test_coding_standards.py#L157-L159

setup.py
```diff
@@ -151,6 +157,7 @@ def _set_builtin(name, value):
     "cmdclass": cmdclass,
     "ext_modules": [udunits_ext],
     "package_data": get_package_data(),
+    "options": {"bdist_wheel": {"py_limited_api": "cp311"}},
```
trexfeathers: Assuming `cp311` is liable to move forward, there should be a line in `test_python_versions()`:

https://github.com/SciTools/cf-units/blob/9e34057547f407df84fc3d5bc742f93831cbbe15/cf_units/tests/test_coding_standards.py#L83-L159

.github/workflows/ci-wheels.yml
ocefpaf: Not really a shifting Window. We will keep cp311 until it breaks, while updating the latest one as soon as a new Python is released.

I'm not familiar with what `test_python_versions` does, but I would still encourage testing things for all supported Python's in the CIs. Only the built wheels will be simplified.

setup.py
```diff
@@ -21,7 +21,12 @@
 COMPILER_DIRECTIVES = {}
 # This Cython macro disables a build warning, obsolete with Cython>=3
 #  see : https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api
-DEFINE_MACROS = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
+DEFINE_MACROS = [
+    ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
+    # 0x030B0000 -> 3.11
+    ("Py_LIMITED_API", "0x030B0000"),
```
trexfeathers: Assuming 3.11 is liable to move forward, there should be a line in `test_python_versions()`:

https://github.com/SciTools/cf-units/blob/9e34057547f407df84fc3d5bc742f93831cbbe15/cf_units/tests/test_coding_standards.py#L83-L159

trexfeathers:
Regarding Tox: we've been worried about it for a while, since tox-conda is unsupported and out-of-date. Unfortunately it has now stopped working. We've been meaning to adopt Pixi anyway so I've spent this afternoon writing a PR for that:

#636 

Hopefully I can get that merged soon and then we can update your PR.

ocefpaf:
I cannot reproduce that failure running `pixi run --environment test-py314 pytest` locally. I'm not sure if it is the compiler version or something else :-/

trexfeathers:
> I cannot reproduce that failure running `pixi run --environment test-py314 pytest` locally. I'm not sure if it is the compiler version or something else :-/

Hey, this is still progress - we have new information! But yeah that's frustrating if it only errors in GHA.

ocefpaf:
> Hey, this is still progress - we have new information! But yeah that's frustrating if it only errors in GHA.

I'll remove the ABI3 from the specialized tests. As long as they build and pass all the tests in the cibuildwheel, that is what really matters. Also, we do need to keep the specialized tests in case folks are building it that way themselves.

There is one expected failure and one, regarding the configuration tests, that is eluding me:

```
XFAIL cf_units/tests/unit/test__udunits2.py::Test_unit::test_offset_by_time_invalid
FAILED cf_units/tests/test_coding_standards.py::test_python_versions - assert 'NAME: "cf-units-py314"' in '# Reference:\n#   - [https://github.com/actions/cache\n#](https://github.com/actions/cache/n#)   - [https://github.com/actions/checkout\n#](https://github.com/actions/checkout/n#)   - [https://github.com/actions/download-artifact\n#](https://github.com/actions/download-artifact/n#)   - [https://github.com/actions/upload-artifact\n#](https://github.com/actions/upload-artifact/n#)   - [https://github.com/prefix-dev/setup-pixi\n\nname:](https://github.com/prefix-dev/setup-pixi/n/nname:) ci-locks\n\non:\n  workflow_dispatch:\n  schedule:\n    - cron: "3 0 * * 1"\n\nconcurrency:\n  group: ${{ github.workflow }}-${{ github.ref }}\n  cancel-in-progress: true\n\njobs:\n  build-locks:\n    name: "Build locks"\n    runs-on: ubuntu-latest\n    defaults:\n      run:\n        shell: bash -l {0}\n    env:\n      SHELLOPTS: "errexit:pipefail"\n      NAME: "cf-units-py313"\n    steps:\n    - name: "Checkout"\n      uses: actions/checkout@v5\n\n    - name: "Pixi setup"\n      uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7\n      with:\n        run-install: false\n\n    - name: "refresh pixi"\n      run: |\n        pixi update --json | pixi exec pixi-diff-to-markdown > diff.md\n        if [ $(wc -c < diff.md) -lt 10 ]; then\n            rm -f diff.md\n        fi\n\n    - name: "refresh locks"\n      if: ${{ hashFiles(\'diff.md\') }}\n      run: |\n        pixi workspace export conda-explicit-spec --envir...ace }}\n\n    - name: "Generate token"\n      uses: actions/create-github-app-token@v2\n      id: generate-token\n      with:\n        app_id: ${{ secrets.AUTH_APP_ID }}\n        private_key: ${{ secrets.AUTH_APP_PRIVATE_KEY }}\n\n    - name: "Push pull-request"\n      id: cpr\n      uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e\n      with:\n        token: ${{ steps.generate-token.outputs.token }}\n        add-paths: ${{ github.workspace }}/requirements/*\n        commit-message: "updated lock files"\n        branch: pixi-auto-update\n        delete-branch: true\n        title: "[cf-units.ci] pixi lock auto-update"\n        body-path: ${{ github.workspace }}/diff.md\n        labels: |\n          Bot\n          New: Pull Request\n\n    - name: "Show pull-request"\n      if: steps.cpr.outputs.pull-request-number != \'\'\n      run: |\n        echo "pull-request #${{ steps.cpr.outputs.pull-request-number }}"\n        echo "pull-request URL ${{ steps.cpr.outputs.pull-request-url }}"\n        echo "pull-request operation [${{ steps.cpr.outputs.pull-request-operation }}]"\n        echo "pull-request SHA ${{ steps.cpr.outputs.pull-request-head-sha }}"\n'
 +  where '# Reference:\n#   - [https://github.com/actions/cache\n#](https://github.com/actions/cache/n#)   - [https://github.com/actions/checkout\n#](https://github.com/actions/checkout/n#)   - [https://github.com/actions/download-artifact\n#](https://github.com/actions/download-artifact/n#)   - [https://github.com/actions/upload-artifact\n#](https://github.com/actions/upload-artifact/n#)   - [https://github.com/prefix-dev/setup-pixi\n\nname:](https://github.com/prefix-dev/setup-pixi/n/nname:) ci-locks\n\non:\n  workflow_dispatch:\n  schedule:\n    - cron: "3 0 * * 1"\n\nconcurrency:\n  group: ${{ github.workflow }}-${{ github.ref }}\n  cancel-in-progress: true\n\njobs:\n  build-locks:\n    name: "Build locks"\n    runs-on: ubuntu-latest\n    defaults:\n      run:\n        shell: bash -l {0}\n    env:\n      SHELLOPTS: "errexit:pipefail"\n      NAME: "cf-units-py313"\n    steps:\n    - name: "Checkout"\n      uses: actions/checkout@v5\n\n    - name: "Pixi setup"\n      uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7\n      with:\n        run-install: false\n\n    - name: "refresh pixi"\n      run: |\n        pixi update --json | pixi exec pixi-diff-to-markdown > diff.md\n        if [ $(wc -c < diff.md) -lt 10 ]; then\n            rm -f diff.md\n        fi\n\n    - name: "refresh locks"\n      if: ${{ hashFiles(\'diff.md\') }}\n      run: |\n        pixi workspace export conda-explicit-spec --envir...ace }}\n\n    - name: "Generate token"\n      uses: actions/create-github-app-token@v2\n      id: generate-token\n      with:\n        app_id: ${{ secrets.AUTH_APP_ID }}\n        private_key: ${{ secrets.AUTH_APP_PRIVATE_KEY }}\n\n    - name: "Push pull-request"\n      id: cpr\n      uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e\n      with:\n        token: ${{ steps.generate-token.outputs.token }}\n        add-paths: ${{ github.workspace }}/requirements/*\n        commit-message: "updated lock files"\n        branch: pixi-auto-update\n        delete-branch: true\n        title: "[cf-units.ci] pixi lock auto-update"\n        body-path: ${{ github.workspace }}/diff.md\n        labels: |\n          Bot\n          New: Pull Request\n\n    - name: "Show pull-request"\n      if: steps.cpr.outputs.pull-request-number != \'\'\n      run: |\n        echo "pull-request #${{ steps.cpr.outputs.pull-request-number }}"\n        echo "pull-request URL ${{ steps.cpr.outputs.pull-request-url }}"\n        echo "pull-request operation [${{ steps.cpr.outputs.pull-request-operation }}]"\n        echo "pull-request SHA ${{ steps.cpr.outputs.pull-request-head-sha }}"\n' = read_text()
 +    where read_text = PosixPath('/home/runner/work/cf-units/cf-units/.github/workflows/ci-locks.yml').read_text
```

trexfeathers:
>There is one expected failure and one, regarding the configuration tests, that is eluding me:

Yeah the output is unclear - I was in a rush! As you might have guessed, this test is to make sure that we change ALL the appropriate files when we advance our supported Python versions; there are so many and it's easy to miss one. In this case it is talking about `ci-locks`, which is still using the `py313` environment while everything else has moved to `py314`.

You just need to change this line:

https://github.com/SciTools/cf-units/blob/19e1055a5be983fdb3afe73a3cad3e52b22193eb/.github/workflows/ci-locks.yml#L28

trexfeathers:
Ugh apparently there are more Python version alignments to make!

This is the offending test:

https://github.com/SciTools/cf-units/blob/19e1055a5be983fdb3afe73a3cad3e52b22193eb/cf_units/tests/test_coding_standards.py#L86

Any ideas for making it less rubbish are welcomed!

... or you can just read the test and then check all the files manually.

ocefpaf:
> Any ideas for making it less rubbish are welcomed!

I don't want to tackle it in the PR, but I can give it a go later. IMO, we can try to make these a "default" latest and forget about them. Some of those can be replaced by pre-commits that updates the `pyproject.toml` based on the `python-required` metadata, etc.

.github/workflows/ci-tests.yml
```diff
@@ -81,7 +82,10 @@ jobs:
         export PATH=${{ matrix.gitpath-prepend }}$PATH
         which git
         git --version
-        pixi run --frozen pytest ${{ matrix.posargs }}
+        pip install -e . --no-build-isolation --no-deps
+        pushd requirements/
+        pixi run --frozen pytest --pyargs cf_units ${{ matrix.posargs }}
```
ocefpaf, line 87: @trexfeathers we need this, or adopt the `src` distribution model, or separate the coverage test into its own github action. However, I would argue that running the tests in the installed library is the correct thing IMO.

trexfeathers: From what I understand, the intention is for Pixi to be 'in charge', so would it actually be appropriate to implement these changes globally, in the Pixi configuration?

https://github.com/SciTools/cf-units/blob/fa88997a9ea3a74cfdc1022f322988b8ead8d806/pyproject.toml#L241-L249

#636 should be seen as a crude first attempt!

trexfeathers: Also I don't understand why are we temporarily jumping into the `requirements/` directory?

setup.py
```diff
@@ -144,10 +157,16 @@ def _set_builtin(name, value):
 
 cmdclass = {"clean_cython": CleanCython, "build_ext": NumpyBuildExt}
 
+if USE_PY_LIMITED_API:
+    SETUP_OPTIONS = {"bdist_wheel": {"py_limited_api": "cp311"}}
```
trexfeathers, line 161: For anyone looking for an explanation in the future: this is not well documented but somewhat summarised in this GitHub issue:

https://github.com/pypa/setuptools/issues/4741

.github/workflows/ci-wheels.yml
```diff
@@ -93,8 +93,10 @@ jobs:
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
       uses: pypa/cibuildwheel@v3.2.1
       env:
-        CIBW_SKIP: "cp39-* pp* *-musllinux*"
+        CIBW_ENVIRONMENT: CF_UNITS_LIMITED_API="1"
+        CIBW_SKIP: "pp* *-musllinux*"
         CIBW_ARCHS: ${{ matrix.arch }}
+        CIBW_BUILD: "cp311-* cp314-*"
```
trexfeathers, line 99: As mentioned in a previous thread - https://github.com/SciTools/cf-units/pull/627#discussion_r2569579169 - I'm keen for something to be added to `test_coding_standards()` to ensure that `cp314` is kept in alignment with all the other places where Python 3.14 is mentioned.

You have experienced yourself how it can be difficult to spot all the places.

Again, sorry for how `test_coding_standards()` is currently written. Hopefully it should be easy enough to add another check to it.

trexfeathers (changes requested):
OK I'm comfortable I understand this now. Just a couple of minor things @ocefpaf. Thanks!

setup.py
ocefpaf: > Also I don't understand why are we temporarily jumping into the requirements/ directory?

Any directory, but the root will do. I was lazy here and copied-and-pasted from my local tests.

> From what I understand, the intention is for Pixi to be 'in charge', so would it actually be appropriate to implement these changes globally, in the Pixi configuration?

That is still a in-place build. While I do believe we could do a global install and test with pixi, I'm not really well versed in it to implement it here. Note that, ideally, we should use the `src` directory model to avoid that entirely. I am not a big fan of the `src` directory model, but it is handy in compiled extensions.

.github/workflows/ci-wheels.yml
ocefpaf: I think we should focus our efforts in eliminating `test_coding_standards` instead of adding a new check. But I'll give it a go...

ocefpaf: Done in the last commit. Let me know if that is OK.

setup.py
ocefpaf: I don't see a clean way to implement a global install and run the tests on it, outside of the local root install to make coverage works on py314. We can:

1. adopt the `src` layout, see https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
1. replace the current in-place build in the pixi config for a `pip install` + directory change, like in the lines above
1. do both in-place and global

Ideally 1 is the cleanest option, but it is a non-trivial amount of work. 3 is to confusing IMO. I would go for 2, but still, I'm not sure it is worth the effort. I do like pixi for managing envs, I do not like how some simple task becomes such a chore and full of repetition.

ocefpaf:
Cool! Another conflict 😭

setup.py
trexfeathers: OK I'm satisfied. Thanks for giving it some thought.

trexfeathers (approved):

PR #636, opened 2025-11-28
Replace Tox with Pixi

trexfeathers:
### Description

tox-conda is out-of-date and unsupported, which is preventing the CI from running. We've known this day was coming for a while, and the loose intention was to replace with Pixi. I'm taking inspiration from:

- GeoVista
- SciTools/mo_pack#143
- SciTools/mo_pack#150


.github/workflows/ci-locks.yml
trexfeathers, line 1: I propose testing this workflow 'live', as we did with mo_pack. Tricky to test before it is part of `main`.

.github/workflows/ci-locks.yml
```diff
@@ -18,83 +18,41 @@ concurrency:
...
-        TOX_INI: ${{ github.workspace }}/tox.ini
-      with:
-        path: ${{ github.workspace }}/.tox
-        key: ${{ runner.os }}-tox-${{ env.ENV_NAME }}-${{ matrix.lock }}-p${{ env.CACHE_PERIOD }}-b${{ env.CACHE_BUILD }}-${{ hashFiles(env.TOX_INI) }}
-
-    - name: "Create ${{ matrix.lock }} conda locks"
-      run: |
-        tox -e ${{ matrix.lock }} -- --platform osx-64 --platform win-64
+        pixi workspace export conda-explicit-spec --environment ${{ env.NAME }} --frozen --ignore-pypi-errors requirements
+        pixi workspace export conda-environment --environment ${{ env.NAME }} requirements/cf-units.yml
```
trexfeathers, line 49: Currently this includes a `pip` section in the file which upsets Readthedocs. Not sure of the correct solution yet; maybe a cleverer use of Pixi environments/features?

See 3033a2a7d0757c8362ae66fd71932df86e981de5 to see what I mean.

Can't see any useful solutions in the [Pixi documentation](https://pixi.sh/dev/reference/cli/pixi/workspace/export/conda-environment/).

bjlittle: Perhaps consider [this pattern](https://github.com/bjlittle/geovista/blob/main/.readthedocs.yml#L11-L30) to use `pixi` natively on `rtd` instead.

.github/workflows/ci-locks.yml
```diff
@@ -18,83 +18,41 @@ concurrency:
...
 
     - name: "Upload lock artifacts"
       uses: actions/upload-artifact@v5
       with:
         name: lock-artifact-${{ matrix.lock }}
-        path: ${{ github.workspace }}/requirements/locks/${{ matrix.lock }}*.lock
+        path: ${{ github.workspace }}/requirements/*
```
trexfeathers: This also needs to include `pixi.lock`.

.github/workflows/ci-locks.yml
```diff
@@ -108,32 +66,30 @@ jobs:
     - name: "Download lock artifacts"
       uses: actions/download-artifact@v6
       with:
-        path: ${{ github.workspace }}/requirements/locks
+        path: ${{ github.workspace }}/requirements
```
trexfeathers: This also needs to include `pixi.lock`.

.github/workflows/ci-tests.yml
```diff
@@ -117,18 +63,25 @@ jobs:
...
+        frozen: true
+        activate-environment: true
+
     - name: "Run ${{ matrix.os }} on ${{ matrix.version}} tests"
       run: |
         export PATH=${{ matrix.gitpath-prepend }}$PATH
         which git
         git --version
-        tox -e ${{ matrix.version }}-${{ matrix.platform }}-test -- ${{ matrix.posargs }}
+        pixi run pytest ${{ matrix.posargs }}
```
bjlittle: Might want to consider using the `pixi run --frozen` here to ensure that the `pixi run` command doesn't attempt to update the `pixi.lock` if it isn't uptodate with the `pyproject.toml` manifest

.github/workflows/ci-locks.yml
trexfeathers: 20c47c5

.github/workflows/ci-locks.yml
trexfeathers: 9b87ab5

.github/workflows/ci-locks.yml
trexfeathers: 9b87ab5

.github/workflows/ci-tests.yml
trexfeathers: 2822a95

.readthedocs.yml
```diff
@@ -3,36 +3,33 @@ version: 2
...
+    install:
+      - pixi install --frozen --environment docs
     post_install:
-      - git stash pop
-
-conda:
-  # NOTE: the activation script for this environment does not get run.
-  #  Therefore, the UDUNITS2 environment variables are not set.
-  #  They are instead set via the RTD dashboard (not yet possible via this
-  #  configuration file - readthedocs/readthedocs.org#6311).
```
trexfeathers, line 28: I was able to remove these from the dashboard since Pixi is taking care of everything ❤️

pyproject.toml
```diff
@@ -191,6 +191,100 @@ exclude = [
...
+dev-py312 = { features = ["dev", "py312"], solve-group = "py312" }
+docs-py312 = { features = ["docs", "py312"], solve-group = "py312" }
+test-py312 = { features = ["test", "dev", "py312"], solve-group = "py312" }
+cf-units-py312 = { features = ["test", "dev", "docs", "py312"], solve-group = "py312" }
+
+py313 = { features = ["py313"], solve-group = "py313" }
+dev-py313 = { features = ["dev", "py313"], solve-group = "py313" }
+docs-py313 = { features = ["docs", "py313"], solve-group = "py313" }
+test-py313 = { features = ["test", "dev", "py313"], solve-group = "py313" }
+cf-units-py313 = { features = ["test", "dev", "docs", "py313"], solve-group = "py313" }
```
trexfeathers: I'm wondering if this manual matrixing is really worth it. Sure it's cool to make maximum use of all the different things Pixi can do, but that doesn't seem justification on its own.

trexfeathers: ... in the meantime it increases complexity which is why I'm getting cold feet.

requirements/README.md
```diff
@@ -1,6 +1,7 @@
 # ⚠️
 
-This directory contains:
+This directory contains auto-generated locked Conda environment files (`*.txt`) for the latest `python` distribution and each `platform` supported by `cf-units`.
```
ESadek-MO, line 3: As I understand it, conda lock is still correct. 

Would pixi lock be less correct (arguably more clear).

pyproject.toml
ESadek-MO: I find it easier to read, honestly? I think that's the only benefit I'm aware of.

ESadek-MO (changes requested):
I've learned as I've gone through this, but as far as my understanding goes this all looks good, and makes sense from the respective docs. 

There are a couple of tiny comments, but nothing of great importance. I'd be happy for this to go as is, if you are!

pyproject.toml
trexfeathers: I went for it b0d38113fe4061202aebe8bb294d7d1b346e9762

requirements/README.md
trexfeathers: Pixi was used to create the file. It is designed to be used by Conda to create an environment (`conda create -n foo --file cf-units-py313_linux-64_conda_spec.txt`). So while we use Pixi to maintain cf-units, someone could make a Conda environment if they wanted.

- conda-lock: a convenience tool which _can_ be used to create such files, but in this case has not been (we used Pixi).
- Pixi lock: that's actually this file: [`pixi.lock`](https://github.com/SciTools/cf-units/blob/e388664ebde36eba927182261a431b13a513bbc2/pixi.lock)

pyproject.toml
trexfeathers: ... aaaand it turns out we were relying on some of it for the CI. I brought it back!

ESadek-MO (approved):
Happy with this! Let's have a live test on main, let's shall!

PR #644, opened 2026-01-16
simplify config a bit

ocefpaf:
@trexfeathers this implements option 2 from https://github.com/SciTools/cf-units/pull/627#discussion_r2698476020
It works locally, let's see if it works in the CI.

@scitools-templating: please no share prompt

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -82,10 +82,7 @@ jobs:
         export PATH=${{ matrix.gitpath-prepend }}$PATH
         which git
         git --version
-        pip install -e . --no-build-isolation --no-deps
-        pushd requirements/
-        pixi run --frozen pytest --pyargs cf_units ${{ matrix.posargs }}
-        popd
+        pixi run --frozen pytest ${{ matrix.posargs }}
 
     - name: "Upload coverage report to Codecov"
       if: contains(matrix.posargs, '--cov')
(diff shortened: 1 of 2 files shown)


trexfeathers:
Templating thing is a work in progress [^1] and not a concern. If everything else passes I will close-and-reopen and it should go away.

[^1]: our attempt to make it possible to manage 10 different repos between 8 different people without losing our sanity

trexfeathers (approved):
Love this, thanks. I had misunderstood and thought you weren't a fan of the idea.

ocefpaf:
> Love this, thanks. I had misunderstood and thought you weren't a fan of the idea.

I have two minds on the src layout vs the flat layout. I prefer the latter for any pure Python project. I'm usually forced into the former when it is a compiled extension for many reasons, here it is the coverage functionality on Python 3.14. With all that said, it felt like walking in circles for simple tasks, so I was not in a hurry to implement. Until a simple solution\* came to mind right as you were merging it 😬.

\* Using a string in the `cmd` pixi config so we can do two commands instead of another node in the build command chain.

trexfeathers:
>I have two minds on the src layout vs the flat layout. I prefer the latter for any pure Python project.

I'm all for aligning with the community; it's easier than forming my own opinions! But we're in no hurry to do any directory re-organising.

ocefpaf:
> I'm all for aligning with the community

Oh. The community is very divided on that one.

PR #685, opened 2026-07-31
Fixing the CI through 3 changes

HGWright:
### Description

0f01174 - updates the manylinux version ref [python-stratify/#439](https://github.com/SciTools/python-stratify/pull/439)

3193cc9 - update the MACOSX_DEPLOYMENT_TARGET to 26 as this is the minimum required by `cf-units` dependencies

18215d4 - added the config for trusted publishing, some config also needs to be done on the pypi and test-pypi pages for this package, and so this PR is blocked until those steps are completed. See docs for more details https://github.com/pypa/gh-action-pypi-publish#usage & https://docs.pypi.org/trusted-publishers/adding-a-publisher/

### DIFF
--- a/.github/workflows/ci-wheels.yml
+++ b/.github/workflows/ci-wheels.yml
@@ -103,7 +103,7 @@ jobs:
         CIBW_ARCHS: ${{ matrix.arch }}
         CIBW_BUILD: "cp311-* cp314-*"
         CIBW_BUILD_FRONTEND: build
-        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux_2_28
         CIBW_BEFORE_BUILD_LINUX: yum install -y udunits2-devel
         CIBW_BEFORE_BUILD_MACOS: brew install udunits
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-wheels.yml
```diff
@@ -176,8 +179,6 @@ jobs:
 
     - uses: pypa/gh-action-pypi-publish@release/v1
       with:
-        user: __token__
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
trexfeathers, line 180: Need to mirror this for the non-test one:

https://github.com/SciTools/cf-units/blob/84a6fd4d44482ebe015e22437bb03f8aabf55f42/.github/workflows/ci-wheels.yml#L204-L205

trexfeathers (changes requested):
Thanks @HGWright! Once change for you.

I've made the requested changes within cf-units PyPI so that's all ready to use.

trexfeathers (approved):

.github/workflows/ci-wheels.yml
HGWright: [4b31349](https://github.com/SciTools/cf-units/pull/685/commits/4b3134962043e502e1a932e8cdff488f78d5b33c)

PR #686, opened 2026-08-19
fix failing ruff checks in precommit

HGWright:
### Description

Ruff un-previewed a couple of new checks and our current licence header needed some config to be accepted by the check. Once this is merged we can update https://github.com/SciTools/cf-units/pull/668, and see if this solves the problem.


HGWright:
invalidated by #687

